PR #2, opened 2017-05-07
Parse input files into a custom AST structure

snim2:
This PR makes the following improvements:

  * The user can now specify two files on the command line (previously only one was expected)
  * A new `treediff::ast` module provides a small implementation of an "index tree"  data structure to represent ASTs.
  * A very simple calculator grammar has been added, to help with unit and integration tests.
  * The user can pass in `-a` or `--ast` to dump out pretty-printed ASTs of the input files to STDOUT
  * The user can pass in `-d FILENAME` or `--dot FILENAME` to dump out AST representations of the input file(s) in [graphviz](http://graphviz.org/) format. As well as adding confidence in the new AST code, this will hopefully useful for debugging later on.

Index trees are a Rust idiom for creating tree / graph structures. The core idea is to create an "arena" which is a vector of nodes in the tree/graph. Each node is a `struct` which contains "pointers" to related nodes (parent, siblings, children). These "pointers" are simply node ids -- i.e. indexes to the vector inside the arena. More complete implementations can be found elsewhere, e.g.:

  * https://github.com/saschagrunert/indextree/
  * https://github.com/SimonSapin/rust-forest/

In the implementation here, nodes contain a number of fields that store details from the parsed input file: indentation level, production name (e.g. `Expr`, `Term`, `Factor`, etc.) and literal values for terminals. More data may be added in later PRs, if necessary.

The index tree implementation here is just sufficient to create and prettty-print an AST, many features are missing (such as iterators, functions to calculate height, size, etc.), and will be added as needed.

### Testing

A number of unit and integration tests have been added in this PR. 

The `-a` option on the command line should print ASTs in exactly the same format as the `lrpar` crate, e.g.:

```sh
$./target/debug/rstreediff tests/one.calc tests/add.calc -a
Expr
 Term
  Factor
   INT 1

Expr
 Term
  Factor
   INT 1
 PLUS +
 Expr
  Term
   Factor
    INT 2

$
```

There are no integration tests for the code which renders a graphviz file, because the [dot](https://docs.rs/dot/0.1.2/dot/) crate renders its output directly to an open file. However, the results for the test files all seem sensible, e.g.:

```sh
$ ./target/debug/rstreediff tests/add.calc -d add.dot tests/Hello.java -d hello.dot
$ dot add.dot -Tpng -oadd.png
$ dot hello.dot -Tpng -ohello.png
```
produces the following output:

**add.png**
add

**hello.png**
hello


Cargo.lock
```diff
@@ -2,7 +2,10 @@
 name = "rstreediff"
 version = "0.1.0"
 dependencies = [
+ "dot 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
```
vext01: Wait, this is a lock file. Should this be in git?

Cargo.toml
```diff
@@ -3,13 +3,24 @@ name = "rstreediff"
 version = "0.1.0"
 authors = ["Sarah Mount <sarah.mount@kcl.ac.uk>"]
```
vext01: Was this based on Geoff's code? Do we need to attribute him?

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
+// Copyright (c) 2017 King's College London
```
vext01: Do we need to have a license on every file? I tend to put one license file in the repo top-level, then you don't need to maintain the copyright year (missing in this file) in multiple places. Not sure what Laurie usually does, perhaps look at his lrtable repo?

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+use std::ops::{Index, IndexMut};
+use std::path::Path;
+
+use dot;
+use lrlex::{build_lex, Lexeme};
+use lrpar::parser;
+use lrtable::{Grammar, Minimiser, TIdx, yacc_to_statetable};
+
+#[derive(Debug)]
+/// Errors raised when parsing a source file.
```
vext01: Why `///` and not `//`? A rust idiom?

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// Parser could not be built by `lrpar`.
+    BrokenParser,
+    /// File contained lexical error and could not be lexed.
+    LexicalError,
+    /// File contained syntax error and could not be parsed.
+    SyntaxError,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd)]
+/// A node identifier for nodes within a particular `Arena`.
```
vext01: Would this be better as a type alias?

https://doc.rust-lang.org/book/type-aliases.html

(Every time I tried to use a type alias, it failed due to inter-module rules).

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+type EdgeId = (usize, usize);
+
+impl fmt::Display for NodeId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:}", self.index)
+    }
+}
+
+/// An arena in which to place `NodeID`-indexed AST nodes.
+pub struct Arena<T, U> {
```
vext01: Perhaps a comment as to what T and U are?

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+
+impl<T, U> Arena<T, U> {
+    /// Create an empty `Arena`.
+    pub fn new() -> Arena<T, U> {
+        Arena { nodes: Vec::new() }
+    }
+
+    /// Create a new node from its data.
+    pub fn new_node(&mut self, data: T, ty: U, indent: u16) -> NodeId {
+        let next_index = self.nodes.len();
```
vext01: Does a vec contain a count, or will the compiler have to emit code to count the items each time?

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// Make one node the child of another.
+    /// Panic if either `NodeId` is not found or this arena is empty.
+    pub fn make_child_of(&mut self, child: NodeId, parent: NodeId) {
+        if child.index >= self.nodes.len() {
+            panic!("Node ID {} not found in arena.", child);
+        }
+        if parent.index >= self.nodes.len() {
+            panic!("Node ID {} not found in arena.", parent);
+        }
+        if self.nodes.is_empty() {
```
vext01: I think this is redundant? One of the two above cases should cause a panic first?

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+            panic!("Arena is empty.")
+        }
+        self[child].parent = Some(parent);
+        match self[parent].first_child {
```
vext01: Not sure what this is all about.

Cargo.toml
snim2: No, not based on Geoff's code, but some of his code may be added later

Cargo.lock
snim2: The consensus seems to be that binaries should commit their lock files, but libraries should not: http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

src/lib/ast.rs
snim2: Triple-slashes are doc comments (like Java), double slashes are "normal" inline comments.

src/lib/ast.rs
snim2: Possibly, I'll have a go...

src/lib/ast.rs
snim2: Here we are making sure that the parent node lists the new node as one of its children. Either as the first and last child (if the parent node does not have any children yet) or as the last child (if it does already have one or more children).

src/lib/ast.rs
snim2: The `lrpar` repo seems to have a license in every file.

src/lib/ast.rs
snim2: I couldn't find a count attribute in the docs, so I think not.

snim2:
OK, I *think* I have responded to everything...

Cargo.lock
ltratt: I must admit, I tend to think that this only makes sense at the point that one makes releases. At the moment if I e.g. decide to use Rust nightly and you use Rust stable -- and the code compiles under both -- I think the `Cargo.lock` can end up looking different for both of us. I might be wrong.

src/lib/ast.rs
ltratt: Yes, it's good practise to have it on every file.

src/lib/ast.rs
```diff
@@ -0,0 +1,617 @@
...
+    /// Parser could not be built by `lrpar`.
+    BrokenParser,
+    /// File contained lexical error and could not be lexed.
+    LexicalError,
+    /// File contained syntax error and could not be parsed.
+    SyntaxError,
+}
+
+/// A node identifier for nodes within a particular `Arena`.
+type NodeId = usize;
```
ltratt: Although you probably realise this, it's worth emphasising (since I was caught out by this) that type aliases are weak. In other words, this just says "you can use NodeId and usize interchangeably". It doesn't actually make a new type called NodeId.

[If you want to make a non-interchangeable type, life gets a little harder. lrtable has examples.]

src/lib/ast.rs
```diff
@@ -0,0 +1,617 @@
...
+
+/// A type to represent edges between nodes within an `Arena`.
+type EdgeId = (usize, usize);
+
+/// An arena in which to place `NodeID`-indexed AST nodes.
+///
+/// `T` and `U` are types given to internal values stored within each node,
+/// these information that the `lrpar` crate will provide, and represent the
```
ltratt: Doesn't parse?

src/lib/ast.rs
```diff
@@ -0,0 +1,617 @@
...
+        next_index
+    }
+
+    /// Return `true` if the arena is empty, `false` otherwise.
+    pub fn is_empty(&self) -> bool {
+        self.nodes.is_empty()
+    }
+
+    /// Make one node the child of another.
+    /// Panic if either `NodeId` is not found.
```
ltratt: In general, the more typical Rust way is to say this returns something like `Error<()>`, so callers can decide whether they want to panic or not. [I'n not saying that's the right design or not here, but it may be worth thinking about.]

src/lib/ast.rs
```diff
@@ -0,0 +1,617 @@
...
+}
+
+impl Arena<String, String> {
+    /// Create a graphviz representation of this arena and write to buffer.
+    pub fn render_dotgraph<W: Write>(&self, buffer: &mut W) -> Result<(), io::Error> {
+        dot::render(self, buffer)
+    }
+}
+
+impl<'a> dot::Labeller<'a, NodeId, EdgeId> for Arena<String, String> {
```
ltratt: I wonder -- and this may be a bad idea! -- if the graphviz stuff should be in a new module? That way you would probably feel more comfortable saying `use dot::{Labeller, ...}` and then not having to qualify `dot::` everywhere.

src/lib/ast.rs
```diff
@@ -0,0 +1,617 @@
...
+        dot::Id::new(format!("N{}", id)).unwrap()
+    }
+
+    fn node_label(&self, id: &NodeId) -> dot::LabelText {
+        let mut label = String::new();
+        label.push_str(self.nodes[*id].ty.as_str());
+        // Only add node.data to the label if it is alphanumeric.
+        let mut is_data_alphanum = true;
+        if self.nodes[*id].data.len() > 0 {
+            for character in self.nodes[*id].data.chars() {
```
ltratt: It might be more idiomatic to do something like:
```rust
if self.nodes[*id].data.chars().iter().find(|c| c.is_alphanumeric()).is_some() {
    label.push(...);
    ...
}
```

src/lib/ast.rs
```diff
@@ -0,0 +1,617 @@
...
+    lexemes.push(Lexeme::new(u16::try_from(usize::from(grm.end_term)).unwrap(),
+                             input.len(),
+                             0));
+
+    // Return parse tree.
+    let pt = match parser::parse::<u16>(&grm, &stable, &lexemes) {
+        Ok(tree) => tree,
+        Err(_) => return Err(ParseError::SyntaxError),
+    };
+    let ast: Arena<String, String> = parse_into_ast(&pt, &grm, &input);
```
ltratt: It would probably be easier to collapse this line into the next.

src/main.rs
```diff
@@ -35,124 +35,209 @@
...
-    if matches.opt_present("h") || matches.free.len() != 1 {
-        usage(prog, "");
-        return;
+    debug!("All arguments: {:?}", args);
+    debug!("Free text arguments: {:?}", matches.free);
+    if matches.opt_present("h") {
+        usage(prog, "", false, options);
+        process::exit(0);
+    }
+    let dump_ast = if matches.opt_present("a") {
```
ltratt: Isn't this equivalent to just `let dump_ast = matches.opt_present("a")`?

ltratt:
I have just done a very quick, very shallow review. Feel free to ignore anything you want, and please don't wait for my approval before merger.

src/lib/ast.rs
snim2: Perhaps I have misunderstood, but I am not sure these do the same thing. We do not want to push `data` to the label at all if it contains any non-alphanum characters...

snim2:
Thanks @ltratt !

src/lib/ast.rs
ltratt: I may have missed off a `!`, but I think the basic idea is sound: you're looking for the first character that fits/doesn't fit a certain criteria. The version I'm suggesting (or some minor variation) can be better optimised by the compiler (and also bails out at the first offending instance, rather than going to the end).

snim2: Hmm, OK I will play around with this...

Cargo.lock
vext01: OK, well I don't know what the lock file is for (dep versions are listed in Cargo.toml). I suppose do what the rust devs do.

Cargo.toml
vext01: ok

src/lib/ast.rs
vext01: OK, well please add the year to the copyright, unify the messages and remember to bump the copyright year if you see on out of date ;)

src/lib/ast.rs
vext01: k

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+    pub fn is_leaf(&self) -> bool {
+        match self.first_child {
+            Some(_) => false,
+            None => true,
+        }
+    }
+
+    /// `true` if this node has no parent.
+    pub fn is_root(&self) -> bool {
+        match self.parent {
```
snim2: Well, yes it is. Also, when you have created a `Node`, and before you have made it a child of another node, it will be a leaf and a root. I am not sure whether it is worth doing something about this...

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+        let mut s = String::new();
+        for _ in 0..self.indent {
+            s.push_str(" ");
+        }
+        write!(f, "{}", s)?; // Indentation.
+        let data_s = format!(" {}", self.data);
+        if data_s.len() == 1 {
+            return write!(f, "{}", self.ty);
+        }
+        write!(f, "{}", self.ty)?;
```
snim2: It swallows the return value from `write`

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    /// Return the ID of the previous sibling of this node, if there is one.
+    pub fn next_sibling(&self) -> Option<NodeId> {
+        self.next_sibling
+    }
+}
+
+impl<T, U> fmt::Display for Node<T, U>
+    where T: fmt::Display,
```
snim2: These say that `T` and `U` must already implement `fmt::Display` otherwise this trait won't work.

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+    let mut parent = vec![None];
+    let mut child_node: NodeId;
+    while !st.is_empty() {
+        let (indent, e) = st.pop().unwrap();
+        match e {
+            &parser::Node::Terminal { lexeme } => {
+                let tid: usize = lexeme.tok_id().try_into().ok().unwrap();
+                let tn = grm.term_name(TIdx::from(tid)).unwrap();
+                let lt = &input[lexeme.start()..lexeme.start() + lexeme.len()];
+                child_node = arena.new_node(lt.to_string(), tn.to_string(), indent);
```
snim2: The `Node` stores the indent level, so we can easily print out a nice AST.

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+    previous_sibling: Option<NodeId>,
+    next_sibling: Option<NodeId>,
+    first_child: Option<NodeId>,
+    last_child: Option<NodeId>,
+    /// The actual data which will be stored within the tree.
+    pub data: T,
+    /// Type of this AST node, derived from grammar e.g. `expr`, `factor`, `term`.
+    pub ty: U,
+    /// Level of indentation, useful for pretty printing, formatting, etc.
+    pub indent: u16,
```
ltratt: I agree that `u16` *might* be too small. I think `u32` is quite big enough for this, and it's better to use less memory than more :)

src/lib/ast.rs
vext01: Here's the disassembly for `len` on a vector:
```
[0x00000b40 0% 5019 (0x0:-1=1)]> pd $r @ sym.__LT_collections..vec..Vec_LT_T_GT__GT_::len::hda194d93e28bba2b                          
/ (fcn) sym.__LT_collections..vec..Vec_LT_T_GT__GT_::len::hda194d93e28bba2b 26                                                        
|   sym.__LT_collections..vec..Vec_LT_T_GT__GT_::len::hda194d93e28bba2b ();                                                           
|           ; var int local_10h @ rbp-0x10                                                                                            
|           ; var int local_8h @ rbp-0x8                                                                                              
|           ; CALL XREF from 0x00001e5c (sym.hello::print_len::h1bd4f4b75b398548)                                                     
|           0x00000b40   *  55             push rbp                                                                                   
|           0x00000b41      4889e5         mov rbp, rsp                                                                               
|           0x00000b44      48897df0       mov qword [rbp - local_10h], rdi                                                           
|           0x00000b48      488b7df0       mov rdi, qword [rbp - local_10h]                                                           
|           0x00000b4c      48897df8       mov qword [rbp - local_8h], rdi                                                            
|           0x00000b50      488b7df8       mov rdi, qword [rbp - local_8h]                                                            
|           0x00000b54      488b4710       mov rax, qword [rdi + 0x10] ; [0x10:8]=0x1003e0003                                         
|           0x00000b58      5d             pop rbp                                                                                    
\           0x00000b59      c3             ret                                                                                        
            0x00000b5a      90             nop                                                                                        
            0x00000b5b      90             nop                                                                                        
            0x00000b5c      90             nop                                                                                        
            0x00000b5d      90             nop                                                                                        
            0x00000b5e      90             nop                                                                                        
            0x00000b5f      90             nop                                                                                        ```

There's no loop, so there is probably a count in the vec, or the compiler figured out the invariant.

Test program I extracted that code from:
```
extern crate rand;

use rand::Rng;

fn print_len(v: Vec<u16>) {
    println!("size={}", v.len());
}

fn main() {
    let mut rng = rand::thread_rng();
    let sz = rng.gen::<u16>();
    let mut v = Vec::new();
    for i in 0..sz {
        v.push(i);
    }
    print_len(v);

}
```

File: src/lib/ast.rs [hunk 22]
```diff
@@ -0,0 +1,623 @@
...
+        }
+        write!(f, "{}", self.ty)?;
+        write!(f, "{}", data_s)
+    }
+}
+
+// Turn a grammar, parser and input string into an AST arena.
+fn parse_into_ast(pt: &parser::Node<u16>, grm: &Grammar, input: &str) -> Arena<String, String> {
+    let mut arena = Arena::new();
+    let mut st = vec![(0, pt)]; // Stack of (indent level, node) pairs
```
snim2: This `parse_into_ast` code was mainly pilfered from the `lrpar` crate, so I'll try to respond to the comments but there may be some bits I don't understand...

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+    // Stack of Option(NodeId)s which are parents of nodes on the `st` stack.
+    // The stack should never be empty, a `None` should be at the bottom of the stack.
+    let mut parent = vec![None];
+    let mut child_node: NodeId;
+    while !st.is_empty() {
+        let (indent, e) = st.pop().unwrap();
+        match e {
+            &parser::Node::Terminal { lexeme } => {
+                let tid: usize = lexeme.tok_id().try_into().ok().unwrap();
```
snim2: Not sure....

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+                let tn = grm.term_name(TIdx::from(tid)).unwrap();
```
ltratt: I disagree: long variable names don't make code easier to read IMHO. The older I get, the shorter my variable names become, and the easier my code is to read.

vext01: Well, i'm all for concise variable names, but i have no idea what `grm` or `lt` is.

ltratt: `grm: &Grammar` isn't obvious enough?

ltratt: [I agree about `lt` though.]

src/lib/ast.rs
vext01: ok

src/lib/ast.rs
vext01: why not unwrap() like you did elsewhere? This way you would get a panic

src/lib/ast.rs
snim2: OK, but according to the docs, `Vec` doesn't seem to have any public attributes: https://doc.rust-lang.org/std/vec/struct.Vec.html

src/lib/ast.rs
```diff
@@ -0,0 +1,623 @@
...
+        while !stack.is_empty() {
+            let id = stack.pop().unwrap();
+            let node = &self.nodes[id.index];
+            if node.is_leaf() {
+                // Terminal node.
+                formatted.push_str(&format!("{}\n", node));
+            } else {
+                // Non-terminal node.
+                formatted.push_str(&format!("{}\n", node));
+                let first_child = node.first_child.unwrap();
```
vext01: I think this is the only comment I am unsure of at the moment

src/lib/ast.rs
snim2: Do we really want a format string to panic though?

src/lib/ast.rs
snim2: In this case I did what I saw in other libraries, I would guess that the reasoning is to have the nodes be a fixed size, whereas with a vector they will grow. It does make traversal a bit more fiddly, but soon-ish I'll add some iterators to help with that.

src/lib/ast.rs
```diff
@@ -364,10 +364,11 @@ fn parse_into_ast(pt: &parser::Node<u16>, grm: &Grammar, input: &str) -> Arena<S
         let (indent, e) = st.pop().unwrap();
         match e {
             &parser::Node::Terminal { lexeme } => {
-                let tid: usize = lexeme.tok_id().try_into().ok().unwrap();
-                let tn = grm.term_name(TIdx::from(tid)).unwrap();
-                let lt = &input[lexeme.start()..lexeme.start() + lexeme.len()];
-                child_node = arena.new_node(lt.to_string(), tn.to_string(), indent);
+                let token_id: usize = lexeme.tok_id().try_into().ok().unwrap();
+                let term_name = grammar.term_name(TIdx::from(token_id)).unwrap();
```
vext01: Laurie objected to `grammar`, but I think it's an improvement tbh.

Split the difference with `grmr`. At least this sounds like `grammar` if pronounced?

snim2:
OK, are we good to squash?

vext01:
go for it

src/lib/ast.rs
ltratt: `grm` is the convention in the `lr*` libraries too FWIW. It would be good to keep it the same here.

snim2:
Splat!

snim2:
And Travis has finished!

PR #3, opened 2017-05-11
Simplify argument parsing.

snim2:
By moving to `docopt`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -35,39 +35,46 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-use std::{env, process};
-use std::fs::File;
-use std::io::{stdout, stderr, Write};
-use std::path::Path;
-
+extern crate docopt;
 #[macro_use]
 extern crate log;
 extern crate env_logger;
+extern crate rustc_serialize;
 
-extern crate getopts;
-use getopts::Options;
+use std::{env, process};
+use std::fs::File;
+use std::io::{stderr, Write};
+use std::path::Path;
+
+use docopt::Docopt;
 
 extern crate treediff;
 use treediff::ast;
 
-fn usage(prog: String, msg: &str, error: bool, options: Options) {
-    let path = Path::new(prog.as_str());
-    let program = match path.file_name() {
-        Some(m) => m.to_str().unwrap(),
-        None => "rstreediff",
-    };
-    let brief = format!("Usage: {} [OPTIONS] <base file> <diff file>", program);
-    let mut output = String::new();
-    if msg.len() > 0 {
-        output.push_str(msg);
-        output.push_str("\n");
-    }
-    output.push_str(options.usage(&brief).as_str());
-    if error {
-        writeln!(stderr(), "{}", output).ok();
-    } else {
-        writeln!(stdout(), "{}", output).ok();
-    }
+const USAGE: &'static str = "
+Usage: rstreediff [options] <base-file> <diff-file>
+       rstreediff [options] <base-file> <diff-file> -d <file> ...
+       rstreediff (--help | --version)
+
+Diff two input files.
+
+Options:
+    -a, --ast         print AST of input files to STDOUT
+    -d, --dot <file>  write out GraphViz representations of the input file(s)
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -198,21 +193,21 @@ fn main() {
...
-    if graphviz.len() > 0 {
-        info!("User wishes to create graphviz files {:?}.", graphviz);
-        write_dotfile_to_disk(&graphviz[0], ast_base);
+    if args.flag_dot.len() > 0 {
+        info!("User wishes to create graphviz files {:?}.", args.flag_dot);
+        write_dotfile_to_disk(&args.flag_dot[0], ast_base);
     }
-    if graphviz.len() > 1 {
-        write_dotfile_to_disk(&graphviz[1], ast_diff);
+    if args.flag_dot.len() > 1 {
```
jaheba: Do you have objections using `!args.flag_dot.is_empty()` instead?

src/main.rs
```diff
@@ -198,21 +193,21 @@ fn main() {
...
-    let ast_diff = parse_file(&matches.free[1], &lex_l_path2, &yacc_y_path2);
+    let ast_base = parse_file(&args.arg_base_file, &lex_l_path1, &yacc_y_path1);
+    let ast_diff = parse_file(&args.arg_diff_file, &lex_l_path2, &yacc_y_path2);
 
     // Dump ASTs to STDOUT, if requested.
-    if dump_ast {
-        println!("{}", ast_base);
-        println!("{}", ast_diff);
+    if args.flag_ast {
+        writeln!(&mut stdout(), "{}", ast_base).ok();
```
jaheba: Is there a reason you are not using `println!`?

src/main.rs
```diff
@@ -128,47 +135,35 @@ fn parse_file(filename: &str,
...
-        Err(f) => {
-            usage(prog, f.to_string().as_str(), true, options);
-            process::exit(1);
-        }
-    };
-    debug!("All arguments: {:?}", args);
-    debug!("Free text arguments: {:?}", matches.free);
-    if matches.opt_present("h") {
-        usage(prog, "", false, options);
+    let argv: Vec<String> = env::args().collect();
```
jaheba: Just a FYI: `let argv: Vec<_>` is normally sufficient. `.collect` needs to now what you want to collect into. The ugly alternative is `.collect::<Vec<_>>()`.

src/main.rs
```diff
@@ -35,39 +35,46 @@
...
-        output.push_str(msg);
-        output.push_str("\n");
-    }
-    output.push_str(options.usage(&brief).as_str());
-    if error {
-        writeln!(stderr(), "{}", output).ok();
-    } else {
-        writeln!(stdout(), "{}", output).ok();
-    }
+const USAGE: &'static str = "
```
jaheba: I'm curious, why have you opted for not using the `docopt!` macro?

src/main.rs
snim2: Not particularly, but it is consistent with the code that writes to stderr.

src/main.rs
snim2: Ah! Because it didn't compile with my version of nightly :)

snim2:
Ready to squash?

src/main.rs
jaheba: Hmm, I personally prefer `println`, because that's what it's for. The actual question is, why doesn't Rust support printing to stderr in a nice way. Defining the macro isn't that [hard](https://stackoverflow.com/questions/27588416/how-to-send-output-to-stderr).

snim2: Ok, I'll change to `println`

snim2:
OK to squash?

jaheba:
There are two  stdout-writes around line 200:

```rust
    // Dump ASTs to STDOUT, if requested.
    if args.flag_ast {
        writeln!(&mut stdout(), "{}", ast_base).ok();
        writeln!(&mut stdout(), "{}", ast_diff).ok();
    }
```

But apart from that, I'm happy :)

snim2:
OK, good to squash now Travis has finished?

snim2:
OK, squashed!

jaheba:
Thanks :)

snim2:
Thanks for reviewing!

PR #4, opened 2017-05-11
New "Action" module

snim2:
An *action* in a tree differ, is some operation that transforms part of an AST (e.g. by deleting a node or inserting a new one).

This PR adds a new `treediff::action` module, which implements the four actions used by GumTree:

  * Delete
  * Insert
  * Move
  * Update

These can be seen in the original [Java code here](https://github.com/GumTreeDiff/gumtree/tree/develop/core/src/main/java/com/github/gumtreediff/actions/model).

It may also be useful to see the GT [code that applies actions to trees](https://github.com/GumTreeDiff/gumtree/blob/297b8f03752cd1a2dc973f637f845b79bcec5741/core/src/main/java/com/github/gumtreediff/actions/ActionUtil.java), and a [concrete implementation of a GT tree](https://github.com/GumTreeDiff/gumtree/blob/297b8f03752cd1a2dc973f637f845b79bcec5741/core/src/main/java/com/github/gumtreediff/tree/Tree.java).

### Design

Each action is implemented as a struct that must implement the `treediff::action::ApplyAction` trait. Only `ApplyAction` and `fmt::Display` have been implemented. 

A core part of the GT design is to create a list of actions and apply them to an AST. Similarly, we can do this with the `treediff::action::ActionList` type, which allows us to treat actions as if they were polymorphic types. e.g.

```rust
let mut arena = create_arena();
let format1 = "Expr +
  INT 1
  Expr *
    INT 3
    INT 4
";
assert_eq!(format1, format!("{}", arena));
// Create action list.
let mut actions: ActionList<String, String> = vec![];
let del = Delete { node: NodeId::new(2) }; // Remove "Expr *".
let ins = Insert {
    data: String::from("2"),
    ty: String::from("INT"),
    indent: 2,
    new_parent: NodeId::new(0),
    nth_child: 1,
};
let upd = Update {
    // Change "+" to "*".
    node: NodeId::new(0),
    value: String::from("*"),
    ty: String::from("Expr"),
};
actions.push(Box::new(del));
actions.push(Box::new(ins));
actions.push(Box::new(upd));
// Apply action list.
actions.apply(&mut arena).unwrap();
let format2 = "Expr *
  INT 1
  INT 2
";
assert_eq!(format2, format!("{}", arena));
```

### Testing

The behaviour of the `rstreediff` binary is unchanged by this PR, so we rely on unit tests.

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -1,0 +1,408 @@
+// Copyright (c) 2017 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file //
+// if one is included with the Software (each a “Larger Work” to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// #![warn(missing_docs)]
+
+use std::clone::Clone;
+use std::fmt;
+
+use ast::{Arena, ArenaError, ArenaResult, NodeId};
+
+/// Apply an action to an AST node.
+pub trait ApplyAction<T: Clone, U: Clone> {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> ArenaResult;
+}
+
+/// A list of actions to be applied.
+///
+/// This type will usually be used by iterating over the list and calling
+/// `apply()` on each element.
+pub type ActionList<T, U> = Vec<Box<ApplyAction<T, U>>>;
+
+/// Delete a node from a given AST.
+pub struct Delete {
+    node: NodeId,
+}
+
+/// Insert a new node into an AST as the `position`th child of an existing node.
+pub struct Insert<T: Clone, U: Clone> {
+    nth_child: u16,
+    data: T,
+    ty: U,
+    indent: u32,
+    new_parent: NodeId,
+}
+
+/// Move a node from one place to another within an AST.
+pub struct Move {
+    from_node: NodeId,
+    parent: NodeId,
+    pos: u16,
+}
+
+/// Update the data inside an AST node.
+pub struct Update<T: Clone, U: Clone> {
+    node: NodeId,
+    value: T,
+    ty: U,
+}
+
+impl fmt::Display for Delete {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "DEL {}", self.node)
+    }
+}
+
+impl<T: fmt::Display + Clone, U: fmt::Display + Clone> fmt::Display for Insert<T, U> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let indent = " ".repeat(self.indent as usize);
+        write!(f,
+               "INS {}{} {} to {} at {}",
+               indent,
+               self.ty,
+               self.data,
+               self.new_parent,
+               self.nth_child)
+    }
+}
+
+impl fmt::Display for Move {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f,
+               "MOV {} to {} at {}",
+               self.from_node,
+               self.parent,
+               self.pos)
+    }
+}
+
+impl<T: fmt::Display + Clone, U: fmt::Display + Clone> fmt::Display for Update<T, U> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "UPD {} to {} {}", self.node, self.ty, self.value)
+    }
+}
+
+impl<T: Clone, U: Clone> ApplyAction<T, U> for Delete {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> ArenaResult {
+        self.node.detach_with_children(arena)
+    }
+}
+
+impl<T: Clone, U: Clone> ApplyAction<T, U> for Insert<T, U> {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> ArenaResult {
+        let mut new_id = arena.new_node(self.data.clone(), self.ty.clone(), self.indent);
+        new_id.make_nth_child_of(self.new_parent, self.nth_child, arena)
+    }
+}
+
+impl<T: Clone, U: Clone> ApplyAction<T, U> for Move {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> ArenaResult {
+        self.from_node
+            .make_nth_child_of(self.parent, self.pos, arena)
+    }
+}
+
+impl<T: Clone, U: Clone> ApplyAction<T, U> for Update<T, U> {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> ArenaResult {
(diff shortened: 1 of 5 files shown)


src/lib/action.rs
```diff
@@ -0,0 +1,411 @@
...
+impl<T: Clone, U: Clone> ApplyAction<T, U> for Move {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> Result<(), ArenaError> {
+        self.from_node
+            .make_nth_child_of(self.parent, self.pos, arena)
+    }
+}
+
+impl<T: Clone, U: Clone> ApplyAction<T, U> for Update<T, U> {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> Result<(), ArenaError> {
+        if !arena.is_in_arena(self.node) {
```
jaheba: not this PR: but shouldn't this method be called `contains`?

src/lib/ast.rs
```diff
@@ -72,19 +70,37 @@ pub enum ParseError {
...
+impl fmt::Display for NodeId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.index)
+    }
+}
 
 /// A type to represent edges between nodes within an `Arena`.
-type EdgeId = (usize, usize);
+///
+/// Private to this module and only used in graphviz-related code.
```
snim2: Oops, that was left over from before the `emitter` module was created!

src/lib/emitters.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    fn node_id(&self, id: &NodeId) -> Id {
+        // Node ids must be unique dot identifiers, be non-empty strings made up
+        // of alphanumeric or underscore characters, not beginning with a digit
+        // (i.e. the regular expression [a-zA-Z_][a-zA-Z_0-9]*).
+        Id::new(format!("N{}", id)).unwrap()
+    }
+
+    fn node_label(&self, id: &NodeId) -> LabelText {
```
snim2: The reference is required by the trait, I think: https://docs.rs/dot/0.1.2/dot/trait.Labeller.html

src/lib/action.rs
```diff
@@ -0,0 +1,411 @@
...
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> Result<(), ArenaError> {
+        let mut new_id = arena.new_node(self.data.clone(), self.ty.clone(), self.indent);
+        new_id.make_nth_child_of(self.new_parent, self.nth_child, arena)
+    }
+}
+
+impl<T: Clone, U: Clone> ApplyAction<T, U> for Move {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> Result<(), ArenaError> {
+        self.from_node
+            .make_nth_child_of(self.parent, self.pos, arena)
```
snim2: Placed there by rustfmt

src/lib/action.rs
snim2: I think this was added in 679d2af0c19d46eab6455154b1723ae9c53e030f so I will change it in this PR.

src/lib/emitters.rs
```diff
@@ -0,0 +1,118 @@
...
+    CouldNotCreateFile,
+    /// Could not write out to a file.
+    CouldNotWriteToFile,
+}
+
+/// Write out a graphviz file (in dot format) to `filepath`.
+pub fn write_dotfile_to_disk(filepath: &str,
+                             arena: &Arena<String, String>)
+                             -> Result<(), EmitterError> {
+    let mut dotfile = match File::create(&filepath) {
```
jaheba: Readding the comment

---

This function body can be rewritten as

```rust
    File::create(&filepath).map_err(|_|EmitterError::CouldNotCreateFile)
        .and_then(|mut dotfile| arena.render_dotgraph(&mut dotfile)
        .map_err(|_|EmitterError::CouldNotWriteToFile))
```

Or using `?` and an thereby implied early return like this:

```rust
    let mut dotfile = File::create(&filepath)
        .map_err(|_|EmitterError::CouldNotCreateFile)?;
    arena.render_dotgraph(&mut dotfile)
        .map_err(|_|EmitterError::CouldNotWriteToFile)
```

Or something inbetween.

src/lib/action.rs
```diff
@@ -0,0 +1,411 @@
...
+// #![warn(missing_docs)]
+
+use std::clone::Clone;
+use std::fmt;
+
+use ast::{Arena, ArenaError, NodeId};
+
+/// Apply an action to an AST node.
+pub trait ApplyAction<T: Clone, U: Clone> {
+    fn apply(&mut self, arena: &mut Arena<T, U>) -> Result<(), ArenaError>;
```
snim2: I'm not sure, there are other places in the code where we return `Result<(), ArenaError>`, so it would probably better to create `ast::ArenaResult` and use that throughout. I'll push a change that implements that and you can see what the impact is.

src/lib/emitters.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn node_label(&self, id: &NodeId) -> LabelText {
+        let label = format!("{} {}", self[*id].ty, self[*id].data);
+        LabelText::LabelStr(label.into())
+    }
+}
+
+impl<'a, T: Clone, U: Clone> GraphWalk<'a, NodeId, EdgeId> for Arena<T, U> {
+    fn nodes(&self) -> Nodes<'a, NodeId> {
+        let mut nodeids: Vec<NodeId> = vec![];
+        for id in 0..self.size() {
```
jaheba: I don't really like the loop here. But the alternative also looks sort of noisy

```rust
        let nodeids = (0..self.size())
            .map(NodeId::new)
            .collect();
```

src/lib/mod.rs
```diff
@@ -45,16 +45,34 @@ extern crate lrlex;
...
 /// Routines are provided to create and iterate over ASTs, and to parse a file
 /// into an AST.
 pub mod ast;
 
-// Re-exported enums and structs.
+/// Emitters generate output for the user in a variety of formats (e.g. JSON, GraphViz).
+pub mod emitters;
+
+// Re-exported enums, structs and types.
+pub use action::Delete;
```
jaheba: Generally these can be written as

```rust
pub use action::{Delete, Insert, Move, ...};
```

src/main.rs
```diff
@@ -77,43 +77,34 @@ struct Args {
     flag_version: bool,
 }
 
-fn write_dotfile_to_disk(filepath: &str, arena: treediff::Arena<String, String>) {
-    let mut dotfile = match File::create(&filepath) {
-        Ok(f) => f,
-        Err(_) => {
-            writeln!(&mut stderr(), "Could not create file {}.", &filepath).ok();
+fn write_dotfile_to_disk(filepath: &str, arena: &ast::Arena<String, String>) {
+    match emitters::write_dotfile_to_disk(filepath, arena) {
```
jaheba: I think I would prefer the code to look a bit like this:

```rust
if let Err(err) = emitters::write_dotfile_to_disk(filepath, arena) {
    let action = match err {
        CouldNotCreateFile => "create",
        CouldNotWriteToFile => "write",
    };
    writeln!(&mut stderr(), "Could not {} to file {}.", action, filepath).ok();
    process::exit(1);
}
```

src/main.rs
```diff
@@ -77,43 +77,34 @@ struct Args {
...
             process::exit(1);
         }
     };
 }
 
-fn parse_file(filename: &str,
-              lexer_path: &str,
-              yacc_path: &str)
-              -> treediff::ast::Arena<String, String> {
+fn parse_file(filename: &str, lexer_path: &str, yacc_path: &str) -> ast::Arena<String, String> {
```
jaheba: Suggestion:

```rust
fn parse_file(filename: &str, lexer_path: &str, yacc_path: &str) -> ast::Arena<String, String> {
    let error_to_str = |err| {
        use ast::ParseError::*;
        match err {
            FileNotFound => "File not found. Check grammar and input files.".into(),
            BrokenLexer => format!("Could not build lexer {}.", lexer_path),
            BrokenParser => format!("Could not build parser {}.", yacc_path),
            LexicalError => format!("Lexical error in {}.", filename),
            SyntaxError => format!("Syntax error in {}.", filename),
            _ => format!("Error parsing {}.", filename),
        }
    };

    ast::parse_file(filename)
        .map_err(error_to_str)
        .map_err(|msg| {
            writeln!(&mut stderr(), "{}", msg).ok();
            process::exit(1);
        })
        .unwrap()
}
```

src/main.rs
snim2: This code in main is very noisy, with all the matches on different error types. However, I'd like to leave this for another PR which would clean up the whole lot in one go.

src/lib/action.rs
jaheba: that tool is a bit strange at times :/

src/main.rs
snim2: Does this really do the same thing as the original code? I think all the error messages that the user sees will be different?

jaheba: I'm fairly sure it does.  Can you explain what do you mean by "all the error messages that the user sees will be different".?

snim2: Ah, I think I misunderstood how the `error_to_str` matcher worked. OK, let me have a go with this...

src/lib/emitters.rs
```diff
@@ -57,16 +57,12 @@ pub enum EmitterError {
...
-        Err(_) => {
-            return Err(EmitterError::CouldNotWriteToFile);
-        }
-    };
+
+    let mut dotfile = File::create(&filepath)
+        .map_err(|_| EmitterError::CouldNotCreateFile)?;
+    arena
+        .render_dotgraph(&mut dotfile)
+        .map_err(|_| EmitterError::CouldNotWriteToFile)?;
```
jaheba:
```rust
arena
        .render_dotgraph(&mut dotfile)
        .map_err(|_| EmitterError::CouldNotWriteToFile)
```

can be your return value

snim2:
OK, how close are we to squashing?

jaheba:
I think there is just one open [comment](https://github.com/softdevteam/rstreediff/pull/4#discussion-diff-116223233R91) left.

Then I also wanted to have a look at the general structure of code, now that I'm a bit more familiar with it.

snim2:
OK, I think I answered the last comment...

jaheba:
Sorry, I meant the loop. It is not a big deal, but I would like to hear your opinion on that.

jaheba:
Looks good to me. Do you want me to have another look at the code or are you happy with how it is now?

snim2:
I'm reasonably happy so far, it's difficult to tell whether APIs are well structured until you write the bits of code that consume them, so I expect I'll be revisiting some of this in later PRs.

jaheba:
Agreed, so let's squash and merge then!

snim2:
Squashed!

snim2:
Many thanks!

PR #5, opened 2017-05-12
Factor out repeated code in main.

snim2:
This is a small improvement in `main()`, but I've had to use the `unreachable!()` macro to avoid errors like this:

```rust
error[E0308]: match arms have incompatible types
   --> src/main.rs:153:22
    |
153 |       let extension2 = match Path::new(&args.arg_diff_file).extension() {
    |  ______________________^
154 | |         Some(ext) => ext.to_str().unwrap(),
155 | |         None => {
156 | |             exit_on_err(format!("Cannot determine file type of {}.", args.arg_diff_file));
157 | |             // unreachable!()
158 | |         }
159 | |     };
    | |_____^ expected &str, found ()
    |
    = note: expected type `&str`
               found type `()`
```

Is there a better way?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -77,6 +77,11 @@ struct Args {
     flag_version: bool,
 }
 
+fn exit_with_message(message: &str) -> ! {
+    writeln!(&mut stderr(), "{}", message).ok();
+    process::exit(1);
+}
+
 fn write_dotfile_to_disk(filepath: &str, arena: &ast::Arena<String, String>) {
     if let Err(err) = emitters::write_dotfile_to_disk(filepath, arena) {
         use emitters::EmitterError::*;
@@ -84,8 +89,7 @@ fn write_dotfile_to_disk(filepath: &str, arena: &ast::Arena<String, String>) {
             CouldNotCreateFile => "create",
             CouldNotWriteToFile => "write to",
         };
-        writeln!(&mut stderr(), "Could not {} file {}.", action, filepath).ok();
-        process::exit(1);
+        exit_with_message(&format!("Could not {} file {}.", action, filepath));
     }
 }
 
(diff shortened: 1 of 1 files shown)


jaheba:
```rust
fn exit_on_err(message: String) {
// to
fn exit_on_err(message: String) -> ! {
```

basically means that exit_on_err does not return

snim2:
Oh, that's really neat!

src/main.rs
```diff
@@ -103,10 +107,7 @@ fn parse_file(filename: &str, lexer_path: &str, yacc_path: &str) -> ast::Arena<S
     };
     ast::parse_file(filename)
         .map_err(error_to_str)
-        .map_err(|msg| {
-                     writeln!(&mut stderr(), "{}", msg).ok();
-                     process::exit(1);
-                 })
+        .map_err(|msg| { exit_on_err(msg); })
```
jaheba: AFAIK this could just be `.map_err(exit_on_err)`.

src/main.rs
```diff
@@ -77,15 +77,19 @@ struct Args {
     flag_version: bool,
 }
 
+fn exit_on_err(message: String) -> ! {
```
jaheba: I think, you normally would just want to have `message: &str` as the parameter, but that requires the function to be called like this

```rust
exit_on_err(&format!("Could not {} file {}.", action, filepath));
```
Which looks a bit more ugly.

snim2:
Well, it seems we can have one of these changes but not both ;)

jaheba:
That's why we can't have nice things :/

I believe using `ref` instead of `&` is generally a bit more defensive, because that closure can take both owned and borrowed values. I can be totally wrong though.

```rust
        .map_err(|ref msg| exit_on_err(msg) )
```

At least I would remove the block (`{ .. }`)  so either

```rust
        .map_err(|ref msg| exit_on_err(msg) )
```
or

```rust
        .map_err(|msg| exit_on_err(&msg) )
```

should do the job.

snim2:
Good to squash?

jaheba:
Looks good to me. Maybe we can find a better name then `exit_on_err`, because for me it implies that the function checks, whether there might an error. Maybe `exit_with_err` is a bit better, but I'm not too sure either.

jaheba:
Even better :) If you squash, I will merge.

snim2:
Squashed!

snim2:
Thanks!

PR #6, opened 2017-05-15
First matcher module PR

snim2:
The PR contains some basic infrastructure for a `matchers` module. A matcher is an algorithm which compares two ASTs, and produces a mapping between similar nodes.

This PR contains some basic structs, constructors, etc. and an implementation of the traits needed to produce a GrahViz description of the two ASTs and their mappings. This can be output on the command line with the new `-m <file>` flag. Mappings between AST nodes are shown with dashed lines, rather than solid, for example:

```sh
$ ./target/debug/rstreediff -m map.dot tests/add.calc tests/mult.calc
$ dot map.dot -Tpng -omap.png
``` 

produces this output:

map

[Needs squashing]

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -344,6 +344,11 @@ impl NodeId {
         NodeId { index: index }
     }
 
+    /// Return the index of this Node Id.
+    pub fn id(&self) -> usize {
+        self.index
+    }
+
     /// Detach this node, leaving its children unaffected.
     ///
     /// Detaching a node makes it inaccessible to other nodes. The node is not
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -12,7 +12,7 @@
 // the Larger Works (as defined below), to deal in both
 //
 // (a) the Software, and
-// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file //
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
 // if one is included with the Software (each a “Larger Work” to which the Software
 // is contributed by such licensors),
 //
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -479,8 +479,8 @@ fn parse_into_ast(pt: &parser::Node<u16>, grm: &Grammar, input: &str) -> Arena<S
     let mut child_node: NodeId;
     while !st.is_empty() {
         let (indent, e) = st.pop().unwrap();
-        match e {
-            &parser::Node::Terminal { lexeme } => {
+        match *e {
+            parser::Node::Terminal { lexeme } => {
                 let token_id: usize = lexeme.tok_id().try_into().ok().unwrap();
                 let term_name = grm.term_name(TIdx::from(token_id)).unwrap();
                 let lexeme_string = &input[lexeme.start()..lexeme.start() + lexeme.len()];
@@ -493,10 +493,10 @@ fn parse_into_ast(pt: &parser::Node<u16>, grm: &Grammar, input: &str) -> Arena<S
                     }
                 };
             }
-            &parser::Node::Nonterminal {
-                 nonterm_idx,
-                 ref nodes,
-             } => {
+            parser::Node::Nonterminal {
+                nonterm_idx,
+                ref nodes,
+            } => {
                 // A non-terminal has no value of its own, but has a node type.
                 child_node = arena.new_node("".to_string(),
                                             grm.nonterm_name(nonterm_idx).unwrap().to_string(),
--- a/src/lib/emitters.rs
+++ b/src/lib/emitters.rs
@@ -41,9 +41,10 @@ use std::borrow::Cow::Owned;
 use std::fs::File;
 use std::io::{Error, Write};
 
-use dot::{Id, Edges, GraphWalk, Labeller, LabelText, Nodes, render};
+use dot::{Id, Edges, GraphWalk, Labeller, LabelText, Nodes, render, Style};
 
 use ast::{Arena, EdgeId, NodeId};
+use matchers::{Mapping, MappingStore};
 
 /// Errors produced in emitting new data.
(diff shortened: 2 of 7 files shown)


src/lib/emitters.rs
```diff
@@ -101,9 +107,86 @@ impl<'a, T: Clone, U: Clone> GraphWalk<'a, NodeId, EdgeId> for Arena<T, U> {
...
+        Owned(nodes)
+    }
+
+    fn edges(&'a self) -> Edges<'a, EdgeId> {
+        let mut edges: Vec<EdgeId> = self.from.get_edges();
+        let size = self.from.size() as usize;
+        for edge in self.to.get_edges() {
+            edges.push((NodeId::new(edge.0.id() + size), NodeId::new(edge.1.id() + size)));
+        }
+        for mapping in &self.mappings {
```
ltratt: Might this be simpler as something like `edges.extend(mappings.map(|x| NodeId::new(m.to.id() + size))`? [The same trick may also work for the loop above, depending on what type `get_edges` returns.]

src/lib/matchers.rs
```diff
@@ -0,0 +1,106 @@
...
+
+impl<T: Clone, U: Clone> MappingStore<T, U> {
+    /// Create a new mapping store.
+    fn new(base: Arena<T, U>, diff: Arena<T, U>) -> MappingStore<T, U> {
+        MappingStore {
+            mappings: vec![],
+            from: base,
+            to: diff,
+        }
+
```
ltratt: Extraneous blank line.

src/main.rs
```diff
@@ -179,4 +194,11 @@ fn main() {
     if args.flag_dot.len() > 1 {
         write_dotfile_to_disk(&args.flag_dot[1], &ast_diff);
     }
+
+    let mapping = matchers::match_trees(ast_base, ast_diff);
+    if !args.flag_map.is_none() {
```
ltratt: Isn't this equivalent to the more obvious `if args.flag_map.is_some()`?

src/lib/ast.rs
```diff
@@ -344,6 +344,12 @@ impl NodeId {
         NodeId { index: index }
     }
 
+    /// Return the index of this Node Id.
+    pub fn id(&self) -> usize {
+        self.index
+    }
+
+
```
ltratt: Extraneous blank line?

src/lib/matchers.rs
```diff
@@ -0,0 +1,106 @@
...
+            to: NodeId::new(to),
+        }
+    }
+}
+
+/// A store of mappings between nodes in different arenas.
+/// Direction is important.
+pub struct MappingStore<T: Clone, U: Clone> {
+    /// Mappings for the stored arenas.
+    pub mappings: Vec<Mapping>,
```
ltratt: Do these things need to be `pub`? I just wondered if the helper methods are enough on their own.

src/lib/emitters.rs
snim2: Interesting, I think it would be neater, but `map` won't work on`self.mappings` because a trait bound is not satisfied.

ltratt:
I'm not quite up to speed on this library yet, so my review is a bit shallow. I did wonder if it was a maybe a bit too Java-ish in places (helper methods that don't do much), but that could just as easily be my ignorance of the overall design.

src/lib/emitters.rs
ltratt: What error do you get? I wonder if it can be easily shut up?

snim2:
It's difficult, when looking at the original GT code, to know what is necessary to the algorithms and what code is there because it is idiomatic to Java. I expect all these APIs will have a lot of churn in the next few PRs.

src/lib/matchers.rs
snim2: Because they are both structures, it seems silly to just have methods which return the whole structure. It might be sensible to restrict access to the structures with iterators and so on, but I'll leave that to another PR.

snim2: `        edges.extend(self.mappings.map(|x| NodeId::new(x.to.id() + size)));` 

gives:

```no method map for Vec<Mapping> in the current scope. note: the method `map` exists but the following trait bounds were not satisfied: `std::vec::Vec<matchers::Mapping> : std::iter::Iterator`, `[matchers::Mapping] : std::iter::Iterator```

**Comment by ltratt** [comment]
Understood.

File: src/lib/matchers.rs (hunk 5 above)
Comment (snim2) [inline src/lib/matchers.rs]: Oh yes, but that gives another error, I think the problem here is that we want to iterate over the vector of mappings, but each mapping is a tuple, which we also need to iterate over:

```
error[E0277]: the trait bound `std::vec::Vec<(ast::NodeId, ast::NodeId)>: std::iter::Extend<ast::NodeId>` is not satisfied
   --> src/lib/emitters.rs:173:15
    |
173 |         edges.extend(self.mappings.iter().map(|x| NodeId::new(x.to.id() + size)));
    |               ^^^^^^ the trait `std::iter::Extend<ast::NodeId>` is not implemented for `std::vec::Vec<(ast::NodeId, ast::NodeId)>`
    |
    = help: the following implementations were found:
              <std::vec::Vec<T> as std::iter::Extend<T>>
              <std::vec::Vec<T> as std::iter::Extend<&'a T>>
```

ltratt: OK, I'm moving this back to the right part of the PR (not sure how it ended up here).

src/lib/emitters.rs
ltratt: Try something like `edges.extend(self.mappings.iter().map(|x| (x.from, NodeId::new(x.to.id() + size))));`

snim2:
OK, good to squash?

snim2:
OK, ready to squash?

ltratt:
Yep please do.

snim2:
OK, squashed

PR #7, opened 2017-05-17
Second matcher module PR

snim2:
This PR adds some more infrastructure needed by the GT matcher algorithm:

  * A priority queue of height-ordered `NodeId`s has been added in the new `hqueue.rs` module.
  * The user can set a number of thresholds from the command line.
  * The API has been updated a little, to match the nomenclature in the GT paper.
  * The `treediff::ast::Arena` type now takes one type variable, rather than two. The type of labels (i.e. the grammar production for each node) is now always `String`.

New CLI options:

```sh
Usage: rstreediff [options] <base-file> <diff-file>
       rstreediff [options] <base-file> <diff-file> -d <file> ...
       rstreediff (--help | --version)

Diff two input files.

Options:
    -a, --ast         print AST of input files to STDOUT.
    --debug LEVEL     debug level used by logger. Valid (case sensitive) values
                      are: Debug, Error, Info, Trace, Warn.
    -d, --dot <file>  write out GraphViz representations of the input file(s).
    -h, --help        print this help menu and exit.
    -m, --map <file>  write out GraphViz representation of the mapping store.
    --max-size VAL    consider subtrees for matching only if they have a size
                      less than VAL.
    --min-dice VAL    set the similarity threshold used for matching ASTs. Two
                      trees are mapped if they are mappable and have a dice
                      coefficient greater than VAL in [0, 1] (default: 0.3).
    --min-height VAL  match only nodes with a height greater than VAL (default: 2).
    -v, --version     print version information and exit.
```

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -53,7 +53,7 @@ use lrtable::{Grammar, Minimiser, TIdx, yacc_to_statetable};
 #[derive(Debug)]
 /// Errors raised when parsing a source file.
 pub enum ParseError {
-    /// Io error returned from standard library routine.
+    /// Io error returned from standard library routines.
     Io(io::Error),
     /// File not found.
     FileNotFound,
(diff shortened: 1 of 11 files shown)


src/lib/ast.rs
```diff
@@ -349,13 +363,28 @@ impl NodeId {
...
+    ///
+    /// The height of a leaf node is 1, the height of a branch node is 1 +
+    /// the height of its tallest child node.
+    pub fn height<T: Clone>(&self, arena: &Arena<T>) -> u32 {
+        if arena[*self].is_leaf() {
+            return 1;
+        }
+        let mut heights: Vec<u32> = self.children(arena)
+            .map(|child| child.height(arena))
+            .collect();
```
ltratt: You might find it more natural to use the `let heights = self.children(...).map(...).collect::<Vec<u32>>()` syntax.

ltratt: [There are some other `collect`s in the PR that this could also be applied to.]

src/lib/hqueue.rs
```diff
@@ -0,0 +1,296 @@
...
+            return None;
+        }
+        Some(self.queue.pop().unwrap().index)
+    }
+
+    /// Push a new node into this priority queue, keeping the queue sorted.
+    ///
+    /// This method has no effect if the new node is already in the queue.
+    /// This method is potentially slow, use `push_many()` to add a number
+    /// of new items to the queue at once.
```
ltratt: How many items deep is the queue likely to get? If it's more than ~16 this method is going to become *really* slow as it's O(n^2) in the worst case. You could make it a linked list at which point it'd become O(n) (but obviously there are other costs with linked lists); or you could replace case 4 with a binary chop which would make it O(n log n). But it's only worth thinking about this if the queue is going to become quite deep.

src/lib/hqueue.rs
```diff
@@ -0,0 +1,296 @@
...
+                if self.queue[index].height() >= height && self.queue[index + 1].height() < height {
+                    self.queue.insert(index, new_node);
+                }
+            }
+        }
+    }
+
+    /// Push a number of new nodes into the queue at once.
+    ///
+    /// This method is a performance improvement on calling `push()` repeatedly.
```
ltratt: Unless you've measured this, I wouldn't be so sure!

src/main.rs
```diff
@@ -146,6 +155,25 @@ fn main() {
         process::exit(0);
     }
 
+    // Set any global constants requested by the user. This should be the ONLY
+    // block of code that mutates these values.
+    let mut matcher_config = matchers::Config::new();
```
ltratt: This is a really long variable name: can we find a nicer shorter one? e.g. something like just `cfg`?

snim2:
The last commit `micro benchmarks for push() and push_many()` is an attempt to respond to your comment `Unless you have measured this..`. This is the result from `cargo bench`:

```
...
test hqueue::tests::bench_push      ... bench:      56,560 ns/iter (+/- 6,824)
test hqueue::tests::bench_push_many ... bench:      66,000 ns/iter (+/- 5,182)
```

So... not quite what I had expected.

ltratt:
It's not entirely surprising, if nothing else because `dedup`probably won't know that the vector is sorted. Anyway, I think we're falling prey to premature optimisation, so let's put this to one side for the time being, and come back to it if/when it turns out to a bottleneck. A lot will depend on how deep these queues get; if they're not very deep, this probably won't be a particularly bad bottleneck.

snim2:
I always assume that the standard library can do better than anything I write, but I guess almost sorted queues are a special case. Is it worth removing `push_many` before we squash?

ltratt:
Yes, personally I would get rid of `push_many`. We can optimise later based on profiling.

[I just read `dedup`'s doc: since it only removes consecutive duplicated elements, my guess about the slowdown was probably wrong. Which just goes to remind me that my guesses about performance are generally terrible ;)]

snim2:
OK, some reasonably big changes in those last commits, but I think this is in decent shape now.

ltratt:
Do you want to use `assert` or `debug_assert`?

ltratt:
Also, in general, I think we prefer tests being in the file they're defined in. It's not a strict rule, so I can be convinced otherwise.

src/lib/hqueue.rs
```diff
@@ -159,7 +159,10 @@ impl HeightQueue {
 
     /// Insert all the children of `parent` into this queue, keeping it sorted.
     pub fn open<T: Clone>(&mut self, parent: &NodeId, arena: Arena<T>) {
-        let children = parent.children(&arena).map(|c| c).collect::<Vec<NodeId>>();
+        let children = parent
+            .children(&arena)
+            .map(|c| c)
```
ltratt: This `map` call isn't doing anything and can be removed, I think?

snim2:
I don't think there are any assertions outside of the tests, and they are conditionally compiled, so I don't think there is a need for `debug_asserts`. In terms of where tests go, the Rust book says unit tests go in the modules themselves and integration tests go in the `tests/` directory, so the `tests/` directory is where the .java and .calc files are used, if that makes sense.

ltratt:
Aha, I didn't spot the asserts were in a test mod. Mea culpa!

Do you want to squash?

snim2:
Splat

PR #8, opened 2017-05-21
Third matcher module PR.

snim2:
This is the third, but not the final PR for implementing a basic tree matcher. This PR introduces more scaffolding and basic algorithms which are necessary to implement the GT algorithm.

In this PR:

  * In `hqueue.rs` various parts of the API have been changed to match the description in the GT paper
  * A number of iterators have been added to the AST module, to traverse subtrees in a variety of orders (breadth-first, pre-order, post-order) which will all needed by algorithms coming in later PRs.
  * In `matchers.rs` the mapping store has been re-written to use hash maps, which matched the GT implementation. This allows constant time look-ups of mappings.
  * In `matchers.rs` a number of "similarity" measures have been implemented, which determine how similar two (previously matched) subtrees are.
  * A small number of other necessary methods have been added to `matchers`.
  * Small bits of code clean up in various places.
  * The `dot` crate, which produces GraphViz output is, unfortunately, not very suitable for dealing with mappings. It cannot create subgraphs, and cannot add arbitrary attributes to nodes and edges. Therefore, I've added a custom function to produce GV output for mapping stores, which gives much more information (e.g. which nodes have been mapped and which not, and at which stage of the GT pipeline the nodes were mapped). This roughly matches the figure in the paper. An example (without many mappings!) is:

map

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -224,7 +224,7 @@ impl<T: fmt::Display + Clone> fmt::Display for Arena<T> {
         while !stack.is_empty() {
             let id = stack.pop().unwrap();
             let node = &self.nodes[id.index];
-            if node.is_leaf() {
+            if id.is_leaf(self) {
                 // Terminal node.
                 formatted.push_str(&format!("{}\n", node));
             } else {
@@ -290,25 +290,6 @@ impl<T: Clone> Node<T> {
         }
     }
 
-    /// `true` if this node has no children.
-    pub fn is_leaf(&self) -> bool {
-        match self.first_child {
-            Some(_) => false,
-            None => true,
-        }
-    }
-
-    /// `true` if this node has no parent.
-    ///
-    /// If a node has been added to the arena without being given a parent, or
-    /// a node has been detached from its parent, this method will return `true`.
-    pub fn is_root(&self) -> bool {
-        match self.parent {
-            Some(_) => false,
-            None => true,
-        }
-    }
-
     /// Return the Id of the parent node, if there is one.
     pub fn parent(&self) -> Option<NodeId> {
         self.parent
@@ -364,12 +345,25 @@ impl NodeId {
         self.index
     }
 
+    /// `true` if this node has no children.
+    pub fn is_leaf<T: Clone>(&self, arena: &Arena<T>) -> bool {
+        arena[*self].first_child.is_none()
+    }
+
+    /// `true` if this node has no parent.
+    ///
+    /// If a node has been added to the arena without being given a parent, or
+    /// a node has been detached from its parent, this method will return `true`.
+    pub fn is_root<T: Clone>(&self, arena: &Arena<T>) -> bool {
+        arena[*self].parent.is_none()
+    }
+
     /// Get the height of this node.
     ///
     /// The height of a leaf node is 1, the height of a branch node is 1 +
     /// the height of its tallest child node.
     pub fn height<T: Clone>(&self, arena: &Arena<T>) -> u32 {
-        if arena[*self].is_leaf() {
+        if self.is_leaf(arena) {
             return 1;
         }
         let mut heights = self.children(arena)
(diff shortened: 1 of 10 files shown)


src/lib/ast.rs
```diff
@@ -364,19 +345,38 @@ impl NodeId {
         self.index
     }
 
+    /// `true` if this node has no children.
+    pub fn is_leaf<T: Clone>(&self, arena: &Arena<T>) -> bool {
+        match arena[*self].first_child {
```
ltratt: Isn't this whole expression equivalent to `arena[*self].first_child.is_none()`?

src/lib/ast.rs
```diff
@@ -364,19 +345,38 @@ impl NodeId {
...
+            None => true,
+        }
+    }
+
+    /// `true` if this node has no parent.
+    ///
+    /// If a node has been added to the arena without being given a parent, or
+    /// a node has been detached from its parent, this method will return `true`.
+    pub fn is_root<T: Clone>(&self, arena: &Arena<T>) -> bool {
+        match arena[*self].parent {
```
ltratt: Similar to above.

src/lib/ast.rs
```diff
@@ -364,19 +345,38 @@ impl NodeId {
...
     ///
     /// The height of a leaf node is 1, the height of a branch node is 1 +
     /// the height of its tallest child node.
     pub fn height<T: Clone>(&self, arena: &Arena<T>) -> u32 {
-        if arena[*self].is_leaf() {
+        if self.is_leaf(arena) {
             return 1;
         }
-        let mut heights = self.children(arena)
+        1 +
```
ltratt: Extraneous blank line?

src/lib/emitters.rs
```diff
@@ -113,84 +114,91 @@ impl RenderDotfile for Arena<String> {
...
+            MappingType::RECOVERY => "[style=dashed, color=green, arrowhead=diamond]",
+        };
+        line = format!("\tFROM{} -> TO{}{};\n", from.id(), to.id(), attrs);
+        digraph.push(line);
     }
+    digraph.push(String::from("}\n"));
+    // Write dotfile to disk.
+    let mut stream = File::create(&filepath)
+        .map_err(|_| EmitterError::CouldNotCreateFile)?;
+    stream
```
ltratt: Extraneous blank line?

src/lib/matchers.rs
```diff
@@ -95,35 +80,264 @@ pub struct MappingStore<T: Clone> {
...
+        self.from_map.contains_key(from)
+    }
+
+    /// `true` if the store has a mapping from a node to `to`.
+    pub fn has_to(&self, to: &NodeId) -> bool {
+        self.to_map.contains_key(to)
+    }
+
+    /// Get the `NodeId` that `to` is mapped from.
+    pub fn get_from(&self, to: &NodeId) -> Option<NodeId> {
```
ltratt: This might be nicer with `self.to_map.get(to).map_or(None, |x| x.0)`?

src/lib/matchers.rs
```diff
@@ -95,35 +80,264 @@ pub struct MappingStore<T: Clone> {
...
+        match self.to_map.get(to) {
+            Some(value) => Some(value.0),
+            None => None,
+        }
+    }
+
+    /// Get the `NodeId` that `from` is mapped to.
+    pub fn get_to(&self, from: &NodeId) -> Option<NodeId> {
+        match self.from_map.get(from) {
```
ltratt: Same as above.

ltratt:
Those are all my comments. If you want to squash, I can merge?

snim2:
Squashed!

ltratt:
I think we should squash away at least the "typo" commit too?

snim2:
Oh, interesting. This is a test case that hasn't yet been used in a test, and was introduced in another PR. So that commit is the only commit in this PR that touches that file, so I left it. WDYT?

ltratt:
Ah, I see. OK, leave it.

PR #9, opened 2017-05-30
First infrastructure PR

snim2:
This PR introduces some of the infrastructure needed to implement the end-to-end diffing tool, in particular the edit script generation algorithm. 

**NEEDS SQUASHING**

Specifically in this PR:

  * A lot of churn on all the core APIs
  * A simple matching algorithm based on Myers longest common subsequence algorithm (this matcher is the default for git-diff).
  * An implementation of the edit script generation algorithm of Chawathe et al. (1996). When this algorithm terminates, it will have transformed the "base" AST into one that is isomorphic to the "diff" AST. It will also have transformed the partial mapping left by the matcher into a total mapping between the new base AST and the original diff AST. (T_1 and T_2 in the paper).
  * Code to draw the result the final mapping left by the action script generator.
  * Command line options to draw the new GraphViz output, choose a matcher and list the available matchers, i.e.:

```bash
$ rstreediff -h

Usage: rstreediff [options] <base-file> <diff-file>
       rstreediff [options] <base-file> <diff-file> -d <file> ...
       rstreediff (--help | --list | --version)

Diff two input files.

Options:
    -a, --ast           print AST of input files to STDOUT.
    --debug LEVEL       debug level used by logger. Valid (case sensitive)
                        values are: Debug, Error, Info, Trace, Warn.
    -d, --dot <file>    write out GraphViz representations of the input file(s).
    --edit <file>       write a GraphViz representation of the mapping store to
                        file, after the edit script has been generated.
    -h, --help          print this help menu and exit.
    -l, --list          print information about the available matchers and exit.
    --map <file>        write out GraphViz representation of the mapping store.
    -m, --matcher ALGO  use a given matching algorithm. Valid (case sensitive)
                        values for ALGO are: GumTree (default), Myers.
    --max-size VAL      consider subtrees for matching only if they have a size
                        less than VAL. GumTree only.
    --min-dice VAL      set the similarity threshold used for matching ASTs.
                        Two trees are mapped if they are mappable and have a
                        dice coefficient greater than VAL in [0, 1] (default:
                        0.3). GumTree only.
    --min-height VAL    match only nodes with a height greater than VAL
                        (default: 2). GumTree only.
    -v, --version       print version information and exit.

$ rstreediff --list 

Myers:
-----
This matcher finds the longest common subsequence between two ASTs. Myers is the
default algorithm used by git-diff.

For more information see Myers (1986) An O(ND) Difference Algorithm and its
Variations.

GumTree
-------
This matcher finds implements the GumTree algorithm, which takes account of MOVE
operations between ASTs.

The GumTree algorithm can be configured with the --min-dice, --max-size and
--min-height switches. See --help for more details.

For more information see Falleri et al. (2014) Find-Grained and Accurate Source
Code Differencing.
```

For example, on these two files:

```bash
$ cat tests/add.calc 
1+2

$ cat tests/mult.calc
3*1+2

$
```

The original mapping from the Myers matcher is:

```bash
$ rstreediff --map map.dot --edit edit.dot -m Myers tests/add.calc tests/mult.calc
```

add_mult_map

And the mapping left by the edit script (new mappings drawn in purple) is:

add_mult_edit

The other way around, i.e.:

```bash
$ rstreediff --map map.dot --edit edit.dot -m Myers tests/mult.calc tests/add.calc 
```
is (note grey nodes are unmapped):
mult_add_map

and the final mapping:
mult_add_edit

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -35,15 +35,17 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-// #![warn(missing_docs)]
+#![warn(missing_docs)]
 
+use std::boxed::Box;
 use std::clone::Clone;
-use std::fmt;
+use std::fmt::{Formatter, Display, Result};
 
 use ast::{Arena, ArenaError, ArenaResult, NodeId};
 
 /// Apply an action to an AST node.
-pub trait ApplyAction<T: Clone> {
+pub trait ApplyAction<T: Clone + Display>: Display {
+    /// Apply an action to an AST.
     fn apply(&mut self, arena: &mut Arena<T>) -> ArenaResult;
 }
 
@@ -51,13 +53,19 @@ pub trait ApplyAction<T: Clone> {
 ///
 /// This type will usually be used by iterating over the list and calling
 /// `apply()` on each element.
-pub type ActionList<T> = Vec<Box<ApplyAction<T>>>;
+pub struct EditScript<T> {
+    actions: Vec<Box<ApplyAction<T>>>,
+}
 
+#[derive(Copy, Clone, Eq, PartialEq)]
 /// Delete a node from a given AST.
+///
+/// It is only valid to delete leaf nodes.
 pub struct Delete {
     node: NodeId,
 }
 
+#[derive(Clone, Eq, PartialEq)]
 /// Insert a new node into an AST as the `position`th child of an existing node.
 pub struct Insert<T: Clone> {
     nth_child: u16,
@@ -67,6 +75,7 @@ pub struct Insert<T: Clone> {
     new_parent: NodeId,
 }
 
+#[derive(Copy, Clone, Eq, PartialEq)]
 /// Move a node from one place to another within an AST.
 pub struct Move {
     from_node: NodeId,
@@ -74,6 +83,7 @@ pub struct Move {
     pos: u16,
 }
 
+#[derive(Clone, Eq, PartialEq)]
 /// Update the data inside an AST node.
 pub struct Update<T: Clone> {
     node: NodeId,
@@ -81,14 +91,56 @@ pub struct Update<T: Clone> {
     label: String,
 }
 
-impl fmt::Display for Delete {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl Delete {
+    /// Create a new `Delete` object.
+    pub fn new(node: NodeId) -> Delete {
+        Delete { node: node }
+    }
+}
+
+impl<T: Display + Clone> Insert<T> {
+    /// Create a new `Insert` object.
+    pub fn new(nth: u16, value: T, label: String, indent: u32, parent: NodeId) -> Insert<T> {
+        Insert {
+            nth_child: nth,
+            value: value,
+            label: label,
+            indent: indent,
+            new_parent: parent,
+        }
+    }
+}
+
+impl Move {
+    /// Create a new `Move` object.
+    pub fn new(from: NodeId, parent: NodeId, position: u16) -> Move {
+        Move {
+            from_node: from,
+            parent: parent,
+            pos: position,
+        }
+    }
+}
+
+impl<T: Display + Clone> Update<T> {
+    /// Create a new `Update` object.
+    pub fn new(node: NodeId, value: T, label: String) -> Update<T> {
+        Update {
+            node: node,
+            value: value,
+            label: label,
+        }
+    }
+}
+
+impl Display for Delete {
+    fn fmt(&self, f: &mut Formatter) -> Result {
         write!(f, "DEL {}", self.node)
     }
 }
 
-impl<T: fmt::Display + Clone> fmt::Display for Insert<T> {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl<T: Display + Clone> Display for Insert<T> {
+    fn fmt(&self, f: &mut Formatter) -> Result {
         let indent = " ".repeat(self.indent as usize);
         write!(f,
                "INS {}{} {} to {} at {}",
@@ -100,8 +152,8 @@ impl<T: fmt::Display + Clone> fmt::Display for Insert<T> {
     }
 }
 
-impl fmt::Display for Move {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl Display for Move {
+    fn fmt(&self, f: &mut Formatter) -> Result {
         write!(f,
                "MOV {} to {} at {}",
                self.from_node,
@@ -110,33 +162,36 @@ impl fmt::Display for Move {
     }
 }
 
-impl<T: fmt::Display + Clone> fmt::Display for Update<T> {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl<T: Display + Clone> Display for Update<T> {
+    fn fmt(&self, f: &mut Formatter) -> Result {
         write!(f, "UPD {} to {} {}", self.node, self.label, self.value)
     }
 }
 
-impl<T: Clone> ApplyAction<T> for Delete {
+impl<T: Clone + Display> ApplyAction<T> for Delete {
     fn apply(&mut self, arena: &mut Arena<T>) -> ArenaResult {
-        self.node.detach_with_children(arena)
+        debug_assert!(self.node.is_leaf(arena),
+                      "Attempt to delete branch node {}",
+                      self.node);
+        self.node.detach(arena)
     }
 }
 
-impl<T: Clone> ApplyAction<T> for Insert<T> {
+impl<T: Clone + Display> ApplyAction<T> for Insert<T> {
     fn apply(&mut self, arena: &mut Arena<T>) -> ArenaResult {
         let mut new_id = arena.new_node(self.value.clone(), self.label.clone(), self.indent);
         new_id.make_nth_child_of(self.new_parent, self.nth_child, arena)
     }
 }
 
-impl<T: Clone> ApplyAction<T> for Move {
+impl<T: Clone + Display> ApplyAction<T> for Move {
     fn apply(&mut self, arena: &mut Arena<T>) -> ArenaResult {
         self.from_node
             .make_nth_child_of(self.parent, self.pos, arena)
     }
 }
 
-impl<T: Clone> ApplyAction<T> for Update<T> {
+impl<T: Clone + Display> ApplyAction<T> for Update<T> {
     fn apply(&mut self, arena: &mut Arena<T>) -> ArenaResult {
         if !arena.contains(self.node) {
             return Err(ArenaError::NodeIdNotFound);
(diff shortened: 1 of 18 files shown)


src/lib/action.rs
```diff
@@ -35,29 +35,37 @@
...
 
+use std::boxed::Box;
 use std::clone::Clone;
 use std::fmt;
 
 use ast::{Arena, ArenaError, ArenaResult, NodeId};
 
 /// Apply an action to an AST node.
-pub trait ApplyAction<T: Clone> {
+pub trait ApplyAction<T: Clone + fmt::Display>: fmt::Display {
```
ltratt: Would it be better to `use std::fmt::Display` above?

src/lib/action.rs
```diff
@@ -35,29 +35,37 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-// #![warn(missing_docs)]
+#![warn(missing_docs)]
```
ltratt: I'm not convinced this is a great setting to leave on: sometimes it just means we end up adding completely pointless documentation.

src/lib/action.rs
```diff
@@ -35,29 +35,37 @@
...
 /// `apply()` on each element.
-pub type ActionList<T> = Vec<Box<ApplyAction<T>>>;
+pub struct EditScript<T> {
+    actions: Vec<Box<ApplyAction<T>>>,
+}
 
+#[derive(Copy, Clone, Eq, PartialEq)]
 /// Delete a node from a given AST.
+///
+/// It is only valid to delete leaf nodes.
```
ltratt: Why is there a struct for it, if it's the only thing which can be deleted? I feel like I might be missing out on something!

src/lib/action.rs
```diff
@@ -147,11 +202,47 @@ impl<T: Clone> ApplyAction<T> for Update<T> {
...
 
-impl<T: Clone> ApplyAction<T> for ActionList<T> {
+impl<T: Clone> Default for EditScript<T> {
+    fn default() -> EditScript<T> {
+        EditScript { actions: vec![] }
+    }
+}
+
+impl<T: Clone + fmt::Display> EditScript<T> {
+    /// Create a new (empty) list of actions.
```
ltratt: Maybe just `Create an empty list of actions`?

src/lib/action.rs
```diff
@@ -147,11 +202,47 @@ impl<T: Clone> ApplyAction<T> for Update<T> {
     }
 }
 
-impl<T: Clone> ApplyAction<T> for ActionList<T> {
+impl<T: Clone> Default for EditScript<T> {
```
ltratt: I'm not sure `Default` is a great name (it's just so generic). Is there a better alternative?

src/lib/ast.rs
```diff
@@ -415,6 +390,31 @@ impl NodeId {
         Ok(())
     }
 
+    /// Make one node the next (i.e. last) child of another.
```
ltratt: I assume `self` becomes the last child of `parent`? Can we clarify?

src/lib/ast.rs
```diff
@@ -415,6 +390,31 @@ impl NodeId {
...
+    pub fn make_child_of<T: Clone>(&self, parent: NodeId, arena: &mut Arena<T>) -> ArenaResult {
+        if self.index >= arena.size() {
```
ltratt: Can this `if` and the next be merged?

src/lib/ast.rs
```diff
@@ -415,6 +390,31 @@ impl NodeId {
...
+            return Err(ArenaError::NodeIdNotFound);
+        }
+        if parent.index >= arena.size() {
+            return Err(ArenaError::NodeIdNotFound);
+        }
+        self.detach(arena)?;
+        arena[*self].parent = Some(parent);
+        match arena[parent].first_child {
```
ltratt: Would this be clearer as:
```rust
if arena[parent].first_child.is_none() {
    arena[parent].first_child = Some(*self);
}
```

src/lib/ast.rs
```diff
@@ -415,6 +390,31 @@ impl NodeId {
...
+        };
+        match arena[parent].last_child {
+            None => arena[parent].last_child = Some(*self),
+            Some(id) => {
+                arena[parent].last_child = Some(*self);
+                arena[id].next_sibling = Some(*self);
+                arena[*self].previous_sibling = Some(id);
+            }
+        };
+        Ok(())
```
ltratt: Why does `Ok` return a (admittedly blank) result?

src/lib/gt_matcher.rs
```diff
@@ -63,11 +65,32 @@ impl Default for GumTreeConfig {
...
+    pub fn new() -> GumTreeConfig {
+        Default::default()
+    }
+}
+
+impl<T: Clone + Display + Eq + 'static> MatchTrees<T> for GumTreeConfig {
+    /// Describe this matcher for the user.
+    fn describe(&self) -> String {
+        let desc = "
+This matcher finds implements the GumTree algorithm, which takes account of MOVE
```
ltratt: Typo "finds implements".

ltratt: Also, should `MOVE` be in capitals?

src/lib/matchers.rs
```diff
@@ -37,9 +37,16 @@
 
 #![warn(missing_docs)]
 
+use std::clone::Clone;
```
ltratt: I think this is in the prelude so doesn't need to be imported?

src/lib/matchers.rs
```diff
@@ -167,15 +231,16 @@ impl<T: Clone> MappingStore<T> {
 
     /// `true` if `from` and `to` may be mapped to one another, `false` otherwise.
     pub fn is_mapping_allowed(&self, from: &NodeId, to: &NodeId) -> bool {
-        self.from[*from].label == self.to[*to].label && !(self.has_from(from) || self.has_to(to))
+        self.from_arena[*from].label == self.to_arena[*to].label &&
+        !(self.contains_from(from) || self.contains_to(to))
```
ltratt: Maybe indent the line a bit more?

src/lib/matchers.rs
```diff
@@ -216,19 +281,264 @@ impl<T: Clone> MappingStore<T> {
...
+        let mut next_from_id = self.from_arena.size();
+        let mut new_mappings = TemporaryMappingStore::new();
+        // Copy current mappings over to new_mappings.
+        for (key, value) in &self.from {
+            new_mappings.push(*key, value.0);
+        }
+        // Combined update, insert, align and move phases.
+        let tmp_to_arena = self.to_arena.clone();
+        for x in root.breadth_first_traversal(&tmp_to_arena) {
+            if x.is_root(&self.to_arena) {
```
ltratt: May be clearer as a `filter` in the `for` loops `iter`?

src/lib/matchers.rs
```diff
@@ -216,19 +281,264 @@ impl<T: Clone> MappingStore<T> {
...
+                      -> Result<(), ArenaError> {
+        debug!("align_children({}, {})", w, x);
+        for child in w.children(&self.from_arena) {
+            from_in_order.remove(&child);
+        }
+        for child in x.children(&self.to_arena) {
+            to_in_order.remove(&child);
+        }
+        let mut s1: Vec<NodeId> = vec![];
+        for child in w.children(&self.from_arena) {
```
ltratt: Can't this be merged with the `for` loop in line 413?

src/lib/matchers.rs
```diff
@@ -216,19 +281,264 @@ impl<T: Clone> MappingStore<T> {
...
+            if new_mappings.contains_from(&child) {
+                let mapped = new_mappings.get_to(&child).unwrap();
+                if x.children(&self.to_arena)
+                       .collect::<Vec<NodeId>>()
+                       .contains(&mapped) {
```
ltratt: It's probably quicker to use `find` (or whatever it's called) rather than `collect` and `contains` (since `collect` is going to malloc memory etc.).

src/lib/matchers.rs
```diff
@@ -216,19 +281,264 @@ impl<T: Clone> MappingStore<T> {
...
+                }
+            }
+        }
+        let s2: Vec<NodeId> = vec![];
+        for child in x.children(&self.to_arena) {
+            if new_mappings.contains_to(&child) {
+                let mapped = new_mappings.get_from(&child).unwrap();
+                if w.children(&self.from_arena)
+                       .collect::<Vec<NodeId>>()
+                       .contains(&mapped) {
```
ltratt: Similar comment to above.

src/main.rs
```diff
@@ -118,10 +135,78 @@ fn consume_emitter_err(res: emitters::EmitterResult, filepath: &str) {
     }
 }
 
+fn consume_edit_script_err(error: &ast::ArenaError) -> ! {
```
ltratt: What does the `!` mean?

src/lib/action.rs
snim2: So that it can implement the `ApplyAction` trait. This is from a previous PR, I think.

src/lib/action.rs
snim2: `Default` is a built in trait

src/main.rs
snim2: It's the divergent type (does not terminate)

src/lib/ast.rs
snim2: Because this function returns an `ArenaResult` which is either a Unit or an `ArenaError`

src/main.rs
ltratt: Interesting -- I did not know that!

src/lib/action.rs
ltratt: Ah, doh, I thought you'd defined a new trait. Sorry!

src/lib/matchers.rs
snim2: I'm not sure, if I say:

```rust
for x in root.breadth_first_traversal(&tmp_to_arena).filter(|n| !n.is_root(&self.to_arena))
```

I get:

```rust
error[E0502]: cannot borrow `self.from_arena` as mutable because `self` is also borrowed as immutable
   --> src/lib/matchers.rs:333:32
    |
313 |         for x in root.breadth_first_traversal(&tmp_to_arena).filter(|n| !n.is_root(&self.to_arena)) {
    |                                                                     ---             ---- previous borrow occurs due to use of `self` in closure
    |                                                                     |
    |                                                                     immutable borrow occurs here
...
333 |                 ins.apply(&mut self.from_arena)?;
    |                                ^^^^^^^^^^^^^^^ mutable borrow occurs here
...
377 |         }
    |         - immutable borrow ends here
```

src/lib/action.rs
ltratt: Makes sense.

src/lib/matchers.rs
ltratt: You might need `iter().filter(...)`?

src/main.rs
snim2: Oddly, the compiler could deduce the ! on `consume_emitter_err` but not this function.

src/lib/matchers.rs
snim2: That also leads to a compiler error and Rust is suggesting implementing a trait from a nightly feature called `core::slice::SliceExt`, which seems like overkill. However, the bigger issue with this algorithm is that it mutates the same tree it iterates over. In this implementation I have `cloned()` the `to_arena` AST. If that turns out to be an expensive operation when we start benchmarking, it may be better to change the underlying AST library to use `Cell` references or similar.

ltratt: OK, let's move on then and ignore my suggestion. We can come back if necessary.

src/lib/matchers.rs
snim2: I've been leaving formatting to `rustfmt`

src/lib/matchers.rs
snim2: I think I have merged these loops, although GH seems to disagree...

snim2:
OK, I think I have responded to the comments so far

ltratt:
Agreed. If you squash, I can merge.

snim2:
OK, squashed...

PR #10, opened 2017-05-31
Second infrastructure PR

snim2:
This PR allows the user to render JSON output. Format matches GumTree output. Default is to produce no output.

e.g. 

```sh
>>> ./target/debug/rstreediff -m Myers -o Json tests/empty.calc tests/one.calc
{
    "matches": [
        {
            "src": 2,
            "dest": 2
        },
        {
            "src": 1,
            "dest": 1
        },
        {
            "src": 3,
            "dest": 3
        },
        {
            "src": 0,
            "dest": 0
        }
    ],
    "actions": [
        {
            "action": "insert",
            "tree": 2,
            "parent": 2,
            "at": 0
        }
    ]
}
```

The output matches GumTree output and is valid JSON:
```sh
>>> ./target/debug/rstreediff -m Myers -o Json tests/Test0.java tests/Test1.java >output.json

>>> import json
>>> with open("output.json") as fp:
...     j = json.loads(fp.read())
...
>>> j['matches']
[{u'dest': 2, u'src': 2}, {u'dest': 1, u'src': 1}, {u'dest': 6, u'src': 12}, {u'dest': 9, u'src': 5}, {u'dest': 10, u'src': 6}, {u'dest': 12, u'src': 8}, {u'dest': 3, u'src': 11}, {u'dest': 8, u'src': 4}, {u'dest': 7, u'src': 3}, {u'dest': 5, u'src': 10}, {u'dest': 11, u'src': 7}, {u'dest': 4, u'src': 9}, {u'dest': 0, u'src': 0}]

>>> len(j['matches'])
13

>>> len(j['actions'])
5
>>>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,7 +42,7 @@ extern crate env_logger;
 extern crate rustc_serialize;
 
 use std::{env, process};
-use std::io::{stderr, Write};
+use std::io::{stderr, stdout, Write};
 use std::path::Path;
 
 use docopt::Docopt;
@@ -81,6 +81,8 @@ Options:
                         0.3). GumTree only.
     --min-height VAL    match only nodes with a height greater than VAL
                         (default: 2). GumTree only.
+    -o, --output FMT    select output format. Valid (case sensitive) values for
+                        FMT are: NoOutput (default), JSON.
     -v, --version       print version information and exit.
 ";
 
@@ -101,6 +103,12 @@ enum Matchers {
     Myers,
 }
 
+#[derive(Debug, Eq, PartialEq, RustcDecodable)]
+enum Output {
+    NoOutput, // Default.
+    JSON,
+}
+
 #[derive(RustcDecodable, Debug)]
 struct Args {
     arg_base_file: String,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/main.rs [hunk 1]
```diff
@@ -101,6 +103,12 @@ enum Matchers {
     Myers,
 }
 
+#[derive(Debug, Eq, PartialEq, RustcDecodable)]
+enum Output {
+    NoOutput, // Default.
+    Json,
```
ltratt: Should this be `JSON` (all caps)?

ltratt:
I only have 1 (very, very trivial) comment.

src/main.rs
snim2: Good question, I can never figure out the Rust conventions, but I think this *can* be all caps, in which case the other enums in main should be too.

ltratt: I don't think the others should be: it's just that JSON really is an acronym in normal day-to-day life.

snim2:
OK, good to squash?

ltratt:
Yep!

snim2:
Done

PR #11, opened 2017-06-02
If a grammar cannot be found, fall back to plain text parser.

snim2:
The intention here is to enable the user to diff files for which we don't yet have a grammar.

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -59,6 +59,8 @@ pub enum ParseError {
     Io(io::Error),
     /// File not found.
     FileNotFound,
+    /// Executable not found (used to determine paths of grammar files).
+    ExeNotFound,
     /// File name had no extension (used to determine which lexer/parser to use).
     NoFileExtension,
     /// Lexer could not be built by `lrlex`.
@@ -676,7 +678,7 @@ fn parse_into_ast(pt: &parser::Node<u16>, grm: &Grammar, input: &str) -> Arena<S
 }
 
 // Read file and return its contents or `ParseError`.
-fn read_file(path: &str) -> Result<String, ParseError> {
+fn read_file(path: &Path) -> Result<String, ParseError> {
     let mut f = match File::open(path) {
         Ok(r) => r,
         Err(e) => return Err(ParseError::Io(e)),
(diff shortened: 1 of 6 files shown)


ltratt:
A dumb question, but just to check: does this definitely work on arbitrary text inputs?

snim2:
I've been trying it out with Rust files and their `.bk` backups and it seems to work OK there.

src/lib/ast.rs
```diff
@@ -695,14 +695,15 @@ pub fn parse_file(input_path: &str) -> Result<Arena<String>, ParseError> {
...
-    let yacc_y_path = format!("grammars/{}.y", extension);
+    let mut lex_l_path = format!("grammars/{}.l", extension);
+    let mut yacc_y_path = format!("grammars/{}.y", extension);
     info!("Using lexer: {} and parser: {} for input: {}",
           &lex_l_path,
           &yacc_y_path,
           input_path);
     if !Path::new(&lex_l_path).exists() || !Path::new(&yacc_y_path).exists() {
-        return Err(ParseError::FileNotFound);
+        lex_l_path = String::from("grammars/txt.l");
```
ltratt: This is a relative path, so doesn't that mean that it won't work if you run the command outside the directory where `grammars` is?

snim2: You are right, it won't work outside the crate root :(

ltratt: I suspect that `std::fs::canonicalize` will be part (but not all) of the solution.

snim2:
This isn't quite what I was expecting, but it seems to work.

snim2:
If I move the binary to `$HOME`, I get this:

```sh
$pwd
/home/snim2
snim2)15:37:57(0)~$ ./rstreediff --debug DEBUG revelation_notes.md revelation_notes.md 
INFO:rstreediff: Selecting the GumTree matching algorithm.
Cannot parse .md files. Falling back to plain text parser.
Cannot parse .md files. Falling back to plain text parser.
INFO:treediff::ast: Using lexer: /home/snim2/Desktop/working/softdev/rstreediff/grammars/txt.l and parser: /home/snim2/Desktop/working/softdev/rstreediff/grammars/txt.y for input: revelation_notes.md
INFO:treediff::ast: Using lexer: /home/snim2/Desktop/working/softdev/rstreediff/grammars/txt.l and parser: /home/snim2/Desktop/working/softdev/rstreediff/grammars/txt.y for input: revelation_notes.md
DEBUG:treediff::matchers: find_pos(1)
DEBUG:treediff::matchers: Edit script: INS paragraph  Parent: document 
...
```

So it works fine. If I move the `rstreediff` directory then that fails. This SO answer seems to suggest that there are better solutions to this problem: https://stackoverflow.com/questions/32748918/is-there-any-way-to-include-binary-or-text-files-in-a-rust-library

ltratt:
In all likelihood it only works because your build directory is still in the same place as when you built it. Put another way: embedding compile-time paths into a binary is a guaranteed way to not have a running system on someone else's computer ;)

Including things in the binary is fine, or just look up something like "how do I know what directory my binary was run from?" The solutions are icky, but they're worth knowing about.

snim2:
Since I put the `grammars/*` file in Cargo.toml I had thought it would packages the files up with the binary, so I'm very confused about this. SO seems to think that this macro can be used to read in files at compile time: https://doc.rust-lang.org/std/macro.include_str.html wdyt?

ltratt:
I'd suggest biting the bullet and working out how to find out what directory the binary is being run in. Slurping things into the binary isn't a scalable solution for big files.

snim2:
Just to be pedantically clear, if I am in directory `/home/flibble/` and the binary is in directory `/home/meep/rstreediff/target/debug/` we should expect the grammars to be stored in `/home/meep/rstreediff/grammars`?

ltratt:
Yes.

snim2:
Ok, I *think* this is what you mean...

ltratt:
Yes, this looks like the right sort of thing. I'm not totally certain if we need all the canonicalisation calls though, but admittedly, they won't hurt too much (though they are quite slow IIRC).

snim2:
Ready to squash?

ltratt:
Yep, please squash.

snim2:
Squashed!

snim2:
Thanks

PR #12, opened 2017-06-02
Fix edit script generator

snim2:
This fixes a bug in the edit script generator which previously caused a crash (some debug code was lookup up nodes in the wrong ASTs). Also, it simplifies an iterator.

Could probably do with squashing.


ltratt:
Agreed, let's squash, then I can merge.

snim2:
Squashed!

PR #13, opened 2017-06-02
Third infrastructure PR

snim2:
This is the third PR to introduce significant new infrastructure, and it will be the last for a while.

This PR introduces coloured diffs on the terminal. At this point, even though the test cases in the repo are very small (in terms of file size) the size of the resulting parse trees is such that the graphviz output and debug output is often too large to look through in detail, this PR is intended to make debugging a little easier.

Because this feature is needed for debugging, the implementation has several deficiencies:

  * Producing a diff is not especially efficient in either time or space.
  * The output cannot be used to produce a patch.
  * Some systems post-process diff output to make it either more accurate or more readable, this implementation does not. Inevitably the output will sometimes be difficult to read.
  * The whole file is printed, rather than just the "context" of the diff.

The colour key is:

  * Deletions -- RED
  * Insertions -- GREEN
  * Updates -- YELLOW
  * Moves -- BLUE (the origin of the move is marked with a circumflex)

## Examples

This is the GT "webdiff" output of the example from the GT paper:

gt_side_by_side

and this is the same output in rstreediff (using the Myers matcher):

java1

Swapping the order of the files:

java2

Some calc files:

calc

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -233,7 +233,12 @@ impl<T: Clone + Display> ApplyAction<T> for Delete {
 
 impl<T: Clone + Display> ApplyAction<T> for Insert<T> {
     fn apply(&mut self, arena: &mut Arena<T>) -> ArenaResult {
-        let mut new_id = arena.new_node(self.value.clone(), self.label.clone(), self.indent);
+        // Set line no, char no, token length to `None`.
+        let mut new_id = arena.new_node(self.value.clone(),
+                                        self.label.clone(),
+                                        self.indent,
+                                        None,
+                                        None);
         new_id.make_nth_child_of(self.new_parent, self.nth_child, arena)
     }
 }
@@ -324,14 +329,14 @@ mod test {
 
     fn create_arena() -> Arena<String> {
         let mut arena = Arena::new();
-        let root = arena.new_node(String::from("+"), String::from("Expr"), 0);
-        let n1 = arena.new_node(String::from("1"), String::from("INT"), 2);
+        let root = arena.new_node(String::from("+"), String::from("Expr"), 0, None, None);
+        let n1 = arena.new_node(String::from("1"), String::from("INT"), 2, None, None);
         n1.make_child_of(root, &mut arena).unwrap();
-        let n2 = arena.new_node(String::from("*"), String::from("Expr"), 2);
+        let n2 = arena.new_node(String::from("*"), String::from("Expr"), 2, None, None);
         n2.make_child_of(root, &mut arena).unwrap();
-        let n3 = arena.new_node(String::from("3"), String::from("INT"), 4);
+        let n3 = arena.new_node(String::from("3"), String::from("INT"), 4, None, None);
         n3.make_child_of(n2, &mut arena).unwrap();
-        let n4 = arena.new_node(String::from("4"), String::from("INT"), 4);
+        let n4 = arena.new_node(String::from("4"), String::from("INT"), 4, None, None);
         n4.make_child_of(n2, &mut arena).unwrap();
         arena
     }
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -43,11 +43,13 @@ use std::fmt::{Formatter, Display, Result};
 
 use ast::{Arena, ArenaError, ArenaResult, NodeId};
 use emitters::RenderJson;
+use matchers::MappingStore;
+use patch::Patch;
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
 /// Type of action.
 ///
 /// Only used where the information related to actions can safely be discarded.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub enum ActionType {
     /// Delete a node in a 'from' arena.
     DELETE,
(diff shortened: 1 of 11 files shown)


src/lib/action.rs
```diff
@@ -43,13 +43,37 @@ use std::fmt::{Formatter, Display, Result};
 
 use ast::{Arena, ArenaError, ArenaResult, NodeId};
 use emitters::RenderJson;
+use matchers::MappingStore;
+use patch::Patch;
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+/// Type of action.
```
ltratt: Shouldn't the comment be before the `derive`?

src/lib/action.rs
```diff
@@ -222,7 +241,63 @@ impl<T: Clone> RenderJson for Update<T> {
     }
 }
 
-impl<T: Clone + Display> ApplyAction<T> for Delete {
+impl<T: Clone + Display + Eq> Patchify<T> for Delete {
+    fn patchify(&self, store: &MappingStore<T>, from: &mut Vec<Patch>, _: &mut Vec<Patch>) {
+        let node = &store.from_arena[self.node];
+        if node.char_no.is_some() && node.token_len.is_some() {
```
ltratt: Does `node.char_no.is_some()` imply `node.token_len.is_some()`?

src/lib/emitters.rs
```diff
@@ -85,6 +94,151 @@ pub fn write_json_to_stream<T: RenderJson, U: RenderJson>(mut stream: Box<Write>
...
+///
+/// This function prints the whole file, rather than just the relevant context.
+/// Patch format is not supported, and hunk headers (e.g. `@@ -1 +1 @@`) are
+/// not printed.
+///
+/// ## Panics
+/// If the terminal foreground cannot be changed or reset by the
+/// [term](https://crates.io/crates/term) crate, this function will panic.
+/// However, `term` is cross-platform so this failure case should be rare.
+pub fn write_colour_diff_to_stdout<T: Clone + Display + Eq>(store: &MappingStore<T>,
```
ltratt: Can we maybe find a snappier name for the function? It's a real mouthful and perhaps a bit over-specific.

src/lib/emitters.rs
```diff
@@ -85,6 +94,151 @@ pub fn write_json_to_stream<T: RenderJson, U: RenderJson>(mut stream: Box<Write>
...
+    let mut to_patches: Vec<Patch> = vec![];
+    script.patchify(store, &mut from_patches, &mut to_patches);
+    // No patches in either AST, so input files must be identical.
+    if from_patches.is_empty() && to_patches.is_empty() {
+        return Ok(());
+    }
+    let to_hunks = hunkify(to_patches);
+    let from_file = read_file(from_path)?;
+    let to_file = read_file(to_path)?;
+    let mut ch = 0;
```
ltratt: This variable is defined a long from where it's used. Maybe put it nearer to line 168?

src/lib/patch.rs
```diff
@@ -0,0 +1,261 @@
...
+#![warn(missing_docs)]
+
+use std::clone::Clone;
+use std::collections::BTreeMap;
+
+use action::ActionType;
+
+// How close do two diffs have to be (in characters) to be merged into one patch?
+// We assume lines are 80 characters long.
+const DIST_THRESHOLD: i64 = 3 * 80;
```
ltratt: Should be a `u64` (or `u`something)?

src/lib/patch.rs
```diff
@@ -0,0 +1,261 @@
...
+            self.start = patch.start;
+        }
+        let p_end = patch.end();
+        if p_end > self.end() {
+            self.length += p_end - self.end();
+        }
+        self.patches.push(patch);
+    }
+
+    fn is_patch_close(&self, patch: &Patch) -> bool {
```
ltratt: Function name is a bit unclear (I had to stop and think).

src/lib/patch.rs
```diff
@@ -0,0 +1,261 @@
...
+        (self.start as i64 - patch.start as i64).abs() <= DIST_THRESHOLD
+    }
+
+    /// Given a character number, return type and length of any patches that
+    /// the character is involved in.
+    ///
+    /// The return value is the action type (`DELETE`, `INSERT`, `MOVE`,
```
ltratt: I don't think we need to list the actiontypes here (it's obvious from the type signature).

src/lib/patch.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// `UPDATE`) and the number of characters from `ch` that are involved in
+    /// that patch.
+    pub fn get_action(&self, ch: usize) -> Option<(ActionType, usize)> {
+        if ch >= self.start && ch < self.end() {
```
ltratt: This feels more like it should be an `assert` than a condition?

ltratt:
Looks good! I wonder if we will want a view where we can turn the red stuff off entirely, but that's something to worry about another day. None of my comments are too major.

src/lib/patch.rs
snim2: No, this is saying "we have some random character from an input file, is the character involved in a patch in this hunk?". The hunk is made up of a number of related patches that are "close" to one another, but the patches are not necessarily adjacent, so there may be gaps between them (the examples above all have one hunk, but there are definitely gaps between the patches!). If the diff output only showed the context of each hunk, we would be able to print out a header above each one, like git-diff does by default.

ltratt: Ah, OK.

snim2:
Is this ready for squashing?

ltratt:
There are still a number of `if node.char_no.is_some() && node.token_len.is_some()`s -- can they all be simplified, or are both parts of the check necessary?

snim2:
OK, all gone now

ltratt:
Then I think we're ready for squashing!

snim2:
Squashed!

PR #14, opened 2017-06-06
Simplify action by relaxing trait bounds.

snim2:
This PR:

  * Removes the requirement for the type variable T to implement the `Display` trait
  * Removes the implementation of `Display` for the types in `action.rs` -- these are not needed by the edit script generator

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -39,7 +39,7 @@
 
 use std::boxed::Box;
 use std::clone::Clone;
-use std::fmt::{Formatter, Display, Result};
+use std::fmt::Debug;
 
 use ast::{Arena, ArenaError, ArenaResult, NodeId};
 use emitters::RenderJson;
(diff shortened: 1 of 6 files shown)


ltratt:
Just to be clear: the things you've changed from `Display` to `Debug` are not meant to be producing format that end users will want to see?

snim2:
No, there are some places where the logger prints `T` typed values for debug output, e.g.:

```rust
debug!("Edit script: INS {} {} Parent: {} {}",
       self.to_arena[x].label, // T type
       self.to_arena[x].value,
       self.from_arena[z].label, // T type
       self.from_arena[z].value);
```

but the terminal output in `emitters::write_diff_to_stdout()` prints its output by reading the input files back in (which is what the GT code does).

ltratt:
So I'm confused. Does that mean we're using the `Debug` trait to mean "pretty print this in a way that end users will want to see"?

snim2:
No, we're just using `Debug` for things like unit tests and logging output.

ltratt:
OK, thanks!

PR #15, opened 2017-06-07
Levenshtein algorithm for string-like types

snim2:
This PR is needed to implement the Zhang-Shasha algorithm, it adds:

  * Levenshtein distance for string-like types
  * Changes LCSS algorithm to use fixed capacity vectors.

No user-facing changes. 

**Needs squashing.**

### DIFF
--- a/src/lib/sequence.rs
+++ b/src/lib/sequence.rs
@@ -37,7 +37,7 @@
 
 #![warn(missing_docs)]
 
-use std::cmp::max;
+use std::cmp::{max, min};
 
 use ast::{Arena, NodeId};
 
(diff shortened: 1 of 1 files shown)


src/lib/sequence.rs
```diff
@@ -88,6 +92,51 @@ pub fn lcss<T: Clone + Eq>(seq1: &[NodeId],
     lcss
 }
 
+/// Three-way minimum.
+fn _min3<T: Ord>(v1: T, v2: T, v3: T) -> T {
```
ltratt: Why `_min3` with an underscore?

src/lib/sequence.rs
```diff
@@ -88,6 +92,51 @@ pub fn lcss<T: Clone + Eq>(seq1: &[NodeId],
...
+}
+
+/// Levenshtein distance with equal weights for deletion, insertion and substitution.
+pub fn levenshtein<T: Eq + PartialEq + ToString>(s1: &T, s2: &T) -> usize {
+    if s1 == s2 {
+        return 0;
+    }
+    let w1 = s1.to_string().chars().collect::<Vec<_>>();
+    let w2 = s2.to_string().chars().collect::<Vec<_>>();
+    if w1.is_empty() {
```
ltratt: It would be quicker to move these checks to before the `collect`.

snim2:
Ready for squashing?

ltratt:
Yep.

snim2:
Splat.

PR #19, opened 2017-06-14
Formatting changes

snim2:
rustfmt has changed recently, and this PR brings the repo up to date...

### DIFF
--- a/src/lib/emitters.rs
+++ b/src/lib/emitters.rs
@@ -90,7 +90,7 @@ pub fn write_json_to_stream<T: RenderJson, U: RenderJson>(mut stream: Box<Write>
         .write_all(format!("{{\n{},\n{}\n}}\n",
                            store.render_json(4),
                            script.render_json(4))
-                           .as_bytes())
+                       .as_bytes())
         .map_err(|_| EmitterError::CouldNotWriteToFile)
 }
 
--- a/src/lib/matchers.rs
+++ b/src/lib/matchers.rs
@@ -349,12 +349,11 @@ impl<T: Clone + Debug + Eq + 'static> MappingStore<T> {
                        self.to_arena[x].value,
                        self.from_arena[z].label,
                        self.from_arena[z].value);
-                w = self.from_arena
-                    .new_node(self.to_arena[x].value.clone(),
-                              self.to_arena[x].label.clone(),
-                              self.to_arena[x].indent,
-                              self.to_arena[x].char_no,
-                              self.to_arena[x].token_len);
+                w = self.from_arena.new_node(self.to_arena[x].value.clone(),
+                                             self.to_arena[x].label.clone(),
+                                             self.to_arena[x].indent,
+                                             self.to_arena[x].char_no,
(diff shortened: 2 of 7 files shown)


snim2:
This has failed because Travis needs to update its version of rustfmt, which it caches. In the branch, I did this by pushing a change to the `.travis.yml` to force-update rustfmt, then I dropped that commit. I guess the same trick is needed here to fix master :(

src/lib/emitters.rs
```diff
@@ -90,7 +90,7 @@ pub fn write_json_to_stream<T: RenderJson, U: RenderJson>(mut stream: Box<Write>
         .write_all(format!("{{\n{},\n{}\n}}\n",
                            store.render_json(4),
                            script.render_json(4))
-                           .as_bytes())
+                       .as_bytes())
```
ltratt: Is this right? The new formatting makes no sense.

src/lib/matchers.rs
```diff
@@ -678,9 +677,9 @@ mod tests {
                             "{\n\"src\": 1,\n\"dest\": 3\n}",
                             "{\n\"src\": 0,\n\"dest\": 0\n}",
                             "{\n\"src\": 2,\n\"dest\": 4\n}"]
-                .iter()
-                .map(|s| String::from(*s))
-                .collect::<Vec<String>>();
+            .iter()
```
ltratt: This indenting makes little sense either.

ltratt:
I wonder if we need to tweak some sort of setting: in a couple of cases rustfmt has really made the indentation more confusing IMHO.

src/lib/matchers.rs
snim2: It isn't easy to see from the context of the diff, which I suspect is wrong (the diff I see in gitk is not the same as the one here on GH). This is the new formatting:
```rust
        let expected = vec!["\"matches\": [",
                            "{\n\"src\": 1,\n\"dest\": 3\n}",
                            "{\n\"src\": 0,\n\"dest\": 0\n}",
                            "{\n\"src\": 2,\n\"dest\": 4\n}"]
            .iter()
            .map(|s| String::from(*s))
            .collect::<Vec<String>>();
```
This is the same text from master branch:
```rust
        let expected = vec!["\"matches\": [",
                            "{\n\"src\": 1,\n\"dest\": 3\n}",
                            "{\n\"src\": 0,\n\"dest\": 0\n}",
                            "{\n\"src\": 2,\n\"dest\": 4\n}"]
                .iter()
                .map(|s| String::from(*s))
                .collect::<Vec<String>>();
```

ltratt: Admittedly, neither indentation level makes any sense to me whatsoever... If we're going to use rustfmt vanilla, we might need to split this into two logical lines, as at the moment it's just too confusing for a human to read IMHO.

snim2:
Needs squashing

ltratt:
OK, although I still don't like the indentation, I guess I can live with it. Squash away!

snim2:
Splat.

PR #24, opened 2017-06-19
Simplifications.

ltratt:
A few simplifications I noticed for rstreediff. Note: all the tests pass, but I don't know if the tests touch all of these code paths. So it might be worth checking that.

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -725,10 +725,7 @@ fn read_file(path: &Path) -> Result<String, ParseError> {
     if !Path::new(path).exists() {
         return Err(ParseError::FileNotFound(path.to_str().unwrap().into()));
     }
-    let mut f = match File::open(path) {
-        Ok(r) => r,
-        Err(e) => return Err(ParseError::Io(e.to_string())),
-    };
+    let mut f = File::open(path).map_err(|e| ParseError::Io(e.to_string()))?;
     let mut s = String::new();
     f.read_to_string(&mut s).unwrap();
     Ok(s)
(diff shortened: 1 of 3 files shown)


snim2:
OK, it will take me a little longer than usual to dig into this one...

snim2:
OK, do you want to bring this branch up to date or shall I?

ltratt:
It's my burden.

ltratt:
OK, rebased. I'm not sure I'm very keen on this line though:

```rust
    let mut f = File::open(path).map_err(|e| ParseError::Io(e.to_string()))?;
```

It feels like we've converted it into a string too soon? However, changing that (or not) would be for another PR.

snim2:
If you leave it as-is, that type is a `std::io::Error` and it seems a bit odd to me to have a `Io(std::io::Error)` in the `ParseError` enumeration. Also, I wanted to derive the `Debug` trait for `ParseError` and that can't be done with the `io::Error` in there.

ltratt:
I think it's fine to just bundle up sub-errors into a big `enum` (I do it in lrtable/cfgrammar using `From`) without munging them. But maybe it's not a good idea here. Whether it is or not, it's not directly related to this PR, so let's discuss it next week, but not hold this PR up for it?

snim2:
I agree. Do you want to squash or shall I merge as-is?

ltratt:
I think the two commits stand nicely alone, so merge as-is from my perspective.

snim2:
Great, thanks!

PR #25, opened 2017-06-21
Test pathological cases in parsing code

snim2:
There is a comment on PR #24 that "all the tests pass, but I don't know if the tests touch all of these code paths". This PR is an attempt to add tests for error types defined in `ast::ParseError`, and to simplify that enumeration.

Can be squashed into one commit.


tests/test_ast.rs
```diff
@@ -234,3 +234,105 @@ fn test_hello_java() {
...
+
+#[test]
+fn test_non_existant_input() {
+    let lex = Path::new("grammars/calc.l");
+    let yacc = Path::new("grammars/calc.y");
+    let ast_result = parse_file("nosuchfileexists.calc", lex, yacc);
+    assert!(ast_result.is_err());
+    match ast_result {
+        Err(ParseError::FileNotFound(s)) => {
+            if s == String::from("nosuchfileexists.calc") {
```
ltratt: Why not just replace this whole `if` with `assert_eq!(s, "nosuchfileexists.calc")` or similar?

tests/test_ast.rs
```diff
@@ -234,3 +234,105 @@ fn test_hello_java() {
...
+
+#[test]
+fn test_non_existant_lex() {
+    let lex = Path::new("grammars/nosuchfileexists.l");
+    let yacc = Path::new("grammars/calc.y");
+    let ast_result = parse_file("tets/one.calc", lex, yacc);
+    assert!(ast_result.is_err());
+    match ast_result {
+        Err(ParseError::FileNotFound(s)) => {
+            if s == String::from("grammars/nosuchfileexists.l") {
```
ltratt: Same comment as above.

tests/test_ast.rs
```diff
@@ -234,3 +234,105 @@ fn test_hello_java() {
...
+        _ => assert!(false),
+    }
+}
+
+#[test]
+fn test_lexical_err() {
+    let lex = Path::new("grammars/calc.l");
+    let yacc = Path::new("grammars/calc.y");
+    let ast_result = parse_file("tests/calc_lexical_err.calc", lex, yacc);
+    assert!(ast_result.is_err());
```
ltratt: This line is redundant given the `match`.

snim2:
OK, ready to squash?

ltratt:
In general, I tend to think `assert!(false)` is less clear than `panic!`. Any thoughts?

snim2:
Well, in a unit test I would expect a test to fail, rather than to cause a panic. If you have lots of failing tests you may still want to run the whole suite to completion.

ltratt:
`panic!` doesn't stop the test suite I'm pretty sure (maybe try to be sure!).

snim2:
It seems you are right about `panic!`, change pushed... Good to squash now?

ltratt:
Yep, looks good to me!

ltratt:
Done?

snim2:
Oops, yes!

PR #27, opened 2017-06-28
Remove lockfile.

snim2:


ltratt:
There is no "+2" button, but I like this very much! Will wait for the tests and then merge.

PR #36, opened 2017-10-03
Add debug (non-whitespace) version of calc grammar 

### LINKED ISSUE #35: Different versions of the grammars
It would be useful (for demonstrations and debugging) to have a version of the `calc` grammar with no whitespace, because the full trees are difficult to read.

snim2:
Intended to simplify demonstration and debugging.

Fixes #35 

With new `--grammar` switch:

  * [map.pdf](https://github.com/softdevteam/diffract/files/1352868/map.pdf)
  * [edit.pdf](https://github.com/softdevteam/diffract/files/1352870/edit.pdf)

Without `--grammar`:

  * [map-full.pdf](https://github.com/softdevteam/diffract/files/1352873/map-full.pdf)
  * [edit-full.pdf](https://github.com/softdevteam/diffract/files/1352875/edit-full.pdf)


src/main.rs
```diff
@@ -69,6 +69,8 @@ Options:
     -d, --dot <file>    write out GraphViz representations of the input file(s).
     --edit <file>       write a GraphViz representation of the mapping store to
                         file, after the edit script has been generated.
+    --grammar           use the 'debug' version of the relevant grammar, if
```
ltratt: This seems like a weird name, and weird thing, for the option to do. Why would I want a debug grammar? I suspect I'm missing out on something!

src/main.rs
```diff
@@ -283,21 +286,60 @@ fn main() {
...
+        canonicalize(dir.join(format!("../../grammars/debug-{}.l", ext1)))
+            .unwrap_or_else(|_| canonicalize(dir.join("../../grammars/txt.l")).unwrap())
+    } else {
+        canonicalize(dir.join(format!("../../grammars/{}.l", ext1)))
+            .unwrap_or_else(|_| canonicalize(dir.join("../../grammars/txt.l")).unwrap())
+    };
+    let yacc_y_path1 = if args.flag_grammar &&
+                          Path::new(&canonicalize(dir.join(format!("../../grammars/debug-{}.y",
+                                                                   ext1)))
+                                        .unwrap())
```
ltratt: The indentation means I'm not sure what this is unwrapping.

src/main.rs
```diff
@@ -283,21 +286,60 @@ fn main() {
...
         }
     };
-    let lex_l_path2 = canonicalize(dir.join(format!("../../grammars/{}.l", ext2)))
-                        .unwrap_or_else(|_| canonicalize(dir.join("../../grammars/txt.l")).unwrap());
-    let yacc_y_path2 = canonicalize(dir.join(format!("../../grammars/{}.y", ext2)))
-                        .unwrap_or_else(|_| canonicalize(dir.join("../../grammars/txt.y")).unwrap());
+    let lex_l_path2 = if args.flag_grammar &&
+                         Path::new(&canonicalize(dir.join(format!("../../grammars/debug-{}.l",
+                                                                  ext2)))
+                                       .unwrap())
```
ltratt: Same comment as above.

snim2:
Ready for squashing?

ltratt:
Obvious question: what's the difference between having an extra flag and the user just passing in a different grammar normally?

snim2:
There is no option to pass in a grammar with Diffract. Instead, Diffract looks at the file extension of your input files, and if it has a grammar that matches it uses that. If not it falls back to the "plain text" grammar.

ltratt:
OK, so perhaps we should think of this as a "grammar override" option (not a debug grammar). Even if diffract has a gramar for file X, you can override it with this; but you could also use it to have diffract parse languages it doesn't natively know about.

snim2:
Well ... that is not how this PR works currently. I've been careful not to assume that the user wishes to diff two files in the same language, so if we want to have a "use my grammar" / "use my lexer" option, then we would potentially have either two or four different switches for lexers / parsers for the two input files.

ltratt:
I understand the PR doesn't have this functionality yet, but I think we need to think about it because it's much more general.

How about if we had an option called (say) `-g` which takes one or two pairs. e.g. `-g java.l,java.y` forces both input files to be parsed as Java but `-g java.l,java.y -g php.l,php.y` forces the "before" file to be parsed as Java and the "after" file as PHP. I'm not saying this is the best possible interface, but I'm trying to think of how we can generalise this in a nice way.

snim2:
I don't especially object to that for debugging functionality, but I think an ordinary end-user should never have to know about parsers and lexers. Do you want this change to go in this PR?

ltratt:
I'm fine with the "choose a grammar by default" which Diffract currently has (indeed, I think it's great!). The suggestion I'm making is for advanced users who have a good reason to use a different lexer/parser.

snim2:
OK, well we might as well use this PR, I'll have a go!

ltratt:
Thanks!

snim2:
OK, this works, modulo the bug I just raised in #39

ltratt:
OK, I think we need to do a bit of squashing then I can merge.

snim2:
Squashed!

PR #45, opened 2017-10-25
Factor out edit script generator

### LINKED ISSUE #43: Factor out edit script generator
...so that Diffract can implement a number of different generation algorithms and the end-user can choose between them.

snim2:
**NEEDS SQUASHING**

This PR factors out the edit script generation code into its own trait, so that we can more easily implement new generators, and the user can choose between them on the command line. Currently, we only have one edit script generation algorithm, but @k1502606 will need this new code in `src/main.rs` soon.

Fixes #43


src/lib/edit_script.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+impl Chawathe96Config {
+    /// Create a new configuration object, with default values.
+    pub fn new() -> Chawathe96Config {
+        Default::default()
+    }
+
+    fn align_children<T: Clone + Debug + Eq + 'static>(&self,
```
ltratt: I think it might be clearer to put the parameters on the next line, rather than having the `<...>` squish them all the way to the RHS of the screen?

src/lib/edit_script.rs
```diff
@@ -0,0 +1,330 @@
...
+        Ok(())
+    }
+
+    /// Find the position of node x in to_arena.
+    fn find_pos<T: Clone + Debug + Eq + 'static>(&self,
+                                                 store: &MappingStore<T>,
+                                                 x: NodeId,
+                                                 new_mappings: &TemporaryMappingStore,
+                                                 to_in_order: &HashSet<NodeId>)
+                                                 -> u16 {
```
ltratt: `u16` seems like quite a small type. Is that deliberate?

src/lib/edit_script.rs
snim2: This is a choice made by rustfmt....

src/lib/edit_script.rs
```diff
@@ -0,0 +1,330 @@
...
+                if x == *child {
+                    return 0;
+                } else {
+                    break;
+                }
+            }
+        }
+        let x_pos = x.get_child_position(&store.to_arena).unwrap();
+        let mut v: Option<NodeId> = None;
+        for (i, child) in x.children(&store.to_arena).enumerate() {
```
ltratt: You might prefer `x.children(&store.to_arena).take(x_pos)` or similar over the `if i >= x_pos` dance inside the loop?

src/lib/edit_script.rs
snim2: This function finds out where a node is in a list of children (i.e. "this is the third child of its parent"), and it seemed unlikely that the number of children would exceed the size of a `u16`, but maybe I'm naive about grammars?

src/lib/edit_script.rs
ltratt: You're not doing much to improve my opinion of rustfmt ;) This is just really painful to read :/ I don't mind the splitting of the args onto multiple lines, but the level of indentation is just too much for me.

src/lib/edit_script.rs
ltratt: Ah, so it's only the direct children? I guess it's OK. But do you ever store this `u16` anywhere? I'm fine with taking a chance if we can cut storage space down, but if it's just a parameter we pass between functions, `u32` might be just as performant in practise? It might be worth knowing that in release mode, rustc's checks on overflow and the like are off.

snim2: If we change this we also need to change some code in `action.rs` which is not changed in this PR (this PR doesn't change any functionality). I'll raise an issue for it and we can think about it another time.

ltratt: Good idea.

src/lib/edit_script.rs
snim2: Hrm.

So this:

```rust
let v = x.children(&store.to_arena).enumerate().filter(|&(i, _)| i >= x_pos).nth(0).map(|(_, c)| c);
```

is apparently different to this:

```rust
let mut v2: Option<NodeId> = None;
for (i, child) in x.children(&store.to_arena).enumerate() {
    if to_in_order.contains(&child) {
        v2 = Some(child);
    }
    if i >= x_pos {
        break;
    }
}
```

ltratt: I was thinking something like:

```rust
for child in x.children(&store.to_arena).take(x_pos) {
    ...
}
```

snim2: That doesn't do the same thing, I think. Let me have another go...

ltratt: Why doesn't it do the same thing? It says "take the first x_pos items", which is what the `break` clause is also enforcing.

snim2:
OK, sorted. Ready to squash?

snim2:
OK, try that

ltratt:
Right, but we really need a test that covers that code, as we've changed its meaning totally twice.

snim2:
Yes, this module isn't well tested by the current test suite, that may have to be a job for tomorrow, but I agree that it's worth putting into this PR

ltratt:
OK, sounds like a plan!

snim2:
This PR is about to split up. This PR will stick with refactoring the edit script generator, which will allow us to easily add new edit script generators. A separate PR will add a fuller suite of integration and unit tests to the edit script generator. Commits to follow...

ltratt:
Sounds like a plan to me!

snim2:
OK, this is ready for re-review / squashing. The Travis failure is happening because we have not brought this branch up to date with `master` (and `master` contains the latest fix for the new `lrpar` API). I'll sort that out at the same time as squashing.

ltratt:
Do you want to rebase before review?

snim2:
Ok, up to date with `master`

snim2:
About to squash (pre-review)

snim2:
Squashed for review (will need more squashing after review).

src/lib/edit_script.rs
```diff
@@ -0,0 +1,297 @@
...
+    let mut to_in_order: HashSet<NodeId> = HashSet::new();
+    let mut new_mappings = TemporaryMappingStore::new();
+    // Copy current mappings over to new_mappings.
+    for (key, value) in &store.from {
+        new_mappings.push(*key, value.0);
+    }
+    // Combined update, insert, align and move phases.
+    let tmp_to_arena = store.to_arena.clone();
+    for x in root.breadth_first_traversal(&tmp_to_arena) {
+        let mut w = root; // Overwritten later.
```
ltratt: The comment is implied by the `mut`, so we can drop the comment?

src/lib/edit_script.rs
```diff
@@ -0,0 +1,297 @@
...
+                       store.from_arena[w].label,
+                       store.from_arena[w].value,
+                       store.from_arena[z].label,
+                       store.from_arena[z].value);
+                mov.apply(&mut store.from_arena)?;
+                script.push(mov);
+            }
+        }
+        align_children(store, w,
+                            x,
```
ltratt: Slightly inconsistent indenting.

src/lib/edit_script.rs
```diff
@@ -0,0 +1,297 @@
...
+    debug!("Append new mappings to store.");
+    for (from, to) in new_mappings.from {
+        if !store.contains_from(&from) {
+            debug!("New mapping due to edit script: {} -> {}", from, to);
+            store.push(from, to, MappingType::EDIT);
+        }
+    }
+    Ok(script)
+}
+
```
ltratt: Double newline.

src/lib/edit_script.rs
```diff
@@ -0,0 +1,297 @@
...
+                  new_mappings: &TemporaryMappingStore,
+                  from_in_order: &mut HashSet<NodeId>,
+                  to_in_order: &mut HashSet<NodeId>)
+                  -> Result<(), ArenaError> {
+    debug!("align_children({}, {})", w, x);
+    for child in x.children(&store.to_arena) {
+        to_in_order.remove(&child);
+    }
+    let mut s1: Vec<NodeId> = vec![];
+    for child in w.children(&store.from_arena) {
```
ltratt: This `for` loop and the one below are identical except for the thing being iterated over. Maybe factor out into a closure?

src/lib/matchers.rs
```diff
@@ -121,6 +121,18 @@ impl TemporaryMappingStore {
...
         self.from.get(from).and_then(|x| Some(*x))
     }
+
+    /// Test whether `n1` is mapped to `n2` in this store.
+    pub fn is_mapped(&self, from: &NodeId, to: &NodeId) -> bool {
+        if !self.contains_from(from) {
+            return false;
+        }
+        let mapped_to = self.get_to(from);
+        if mapped_to.is_none() {
```
ltratt: This feels like it might be neater as something like:

```rust
mapped_to.map_or(false, |x| x == to)
```
?

(same with the similar statement below)

src/lib/edit_script.rs
```diff
@@ -134,43 +154,57 @@ impl Chawathe96Config {
...
+        // 4. Let u be the partner of v in T_1.
+        let u_opt = new_mappings.get_from(&v.unwrap());
+        assert!(u_opt.is_some(), "find_pos() could not find a partner for v");
+        let u = u_opt.unwrap();
+        // 5. Suppose u is the i'th child of its parent that is marked
+        // "in order". Return i + 1.
+        let u_p = store.from_arena[u].parent().unwrap();
+        let u_pos = u.get_child_position(&store.from_arena).unwrap();
+        let ret = u_p.children(&store.from_arena)
+            .take(u_pos)
```
ltratt: Indentation doesn't line up.

ltratt:
I only have those fairly shallow comments.

src/lib/edit_script.rs
snim2: `rustfmt` doesn't change this file, and that `.take` is 4 spaces to the right of the `l` in the `let`... ?

ltratt:
Looks good! Squash and then I can merge?

snim2:
OK, squashed...

PR #53, opened 2017-11-24
Clippy suggestions

snim2:
Minor improvements suggested by Clippy. I'm never really sure whether it's sensible to keep these commits separate (for merging, clearly it's better to separate them for reviewing), but potentially they could be collapsed into a single commit.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -56,7 +56,7 @@ use diffract::gt_matcher;
 use diffract::myers_matcher;
 use diffract::matchers::MatchTrees;
 
-const USAGE: &'static str = "
+const USAGE: &str = "
 Usage: diffract [options] <base-file> <diff-file>
        diffract [options] <base-file> <diff-file> -d FILE ...
        diffract [options] <base-file> <diff-file> -g FILE ...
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -361,6 +349,27 @@ fn main() {
             exit_with_message(&format!("Requested parser {:?} does not exist.", parser2));
         }
     }
+    (lexer1, parser1, lexer2, parser2)
+}
+
+fn main() {
+    let argv: Vec<_> = env::args().collect();
+    let args: Args = Docopt::new(USAGE)
+        .and_then(|d| d.argv(argv).decode())
```
ltratt: Not indented properly?

ltratt:
Is this ready for squashing? [I think so, but am not sure...]

snim2:
If you are happy then yes. Do you think it is sensible to squash a PR like this into a single commit (I'm never sure)?

ltratt:
I don't think there's any one-size-fits-all rule for this sort of squashing. I like to see logically distinct commits with messages that explain clearly what's going on; sometimes there's only one such commit; sometimes there are several. So up to you what you think best in this case.

snim2:
OK, squashed!

PR #58, opened 2017-12-04
Added new matcher and action operation

k1502606:


src/lib/action.rs
```diff
@@ -59,6 +59,10 @@ pub enum ActionType {
     MOVE,
     /// Update the text or label associated with a node.
     UPDATE,
+    /// Copy the text or label associated with a node or subtree.
+    COPY,
+    /// GLUE takes a subtree/node from tree1 and inserts it into tree2
```
snim2: OK, but maybe say `from_node` and `parent` rather than `tree1` and `tree2` (to fit with your code in the struct)

src/lib/action.rs
```diff
@@ -59,6 +59,10 @@ pub enum ActionType {
     MOVE,
     /// Update the text or label associated with a node.
     UPDATE,
+    /// Copy the text or label associated with a node or subtree.
```
snim2: I think this copies the whole nodes / subtrees, right? Not just the labels and values (this is maybe a little pedantic, because the paper suggests that nodes contain a little less information than the Diffract implementation puts inside them)

src/lib/action.rs
```diff
@@ -267,6 +335,40 @@ impl<T: Clone + Debug + Eq + 'static> Patchify<T> for Update<T> {
     }
 }
 
+impl<T: Clone + Debug + Eq + 'static> Patchify<T> for Copy{
+    fn patchify(&self, store: &MappingStore<T>, from: &mut Vec<Patch>, to: &mut Vec<Patch>) {
```
snim2: OK, the `patchify` trait looks OK. This is something for us to think about later down the line, because `patchify` takes a `MappingStore` as an argument. However, let's leave the code you have for now and we'll think about it again when we get to producing diffs.

src/lib/ast.rs
```diff
@@ -306,6 +306,7 @@ pub struct Node<T: Clone> {
     pub token_len: Option<usize>,
 }
 
+
```
snim2: Remove blank line.

src/lib/ast.rs
```diff
@@ -473,6 +474,46 @@ impl NodeId {
...
+        let _new_node_made = _arena.new_node(current_node.value,
+                        current_node.label,
+                        current_node.indent,
+                        current_node.col_no,
+                        current_node.line_no,
+                        current_node.char_no,
+                        current_node.token_len,
+                        );
+        // Collect all the children 
+        //let ids = current_node.NodeId.children(_arena).collect::<Vec<NodeId>>();//current_node.id.children(arena).collect::<Vec<NodeId>>();
```
snim2: Remove commented code.

src/lib/ast.rs
```diff
@@ -473,6 +474,46 @@ impl NodeId {
...
+                                                );
+            _new_child_node.make_nth_child_of(_new_node_made,position,_arena);
+            if !id.is_leaf(_arena) {
+                _new_child_node.copy_subtree(id, counter,_arena);
+            }
+            counter=counter+1;
+        }
+
+        self.make_nth_child_of( _new_node_made, position, _arena)
+       // Ok(())
```
snim2: Remove commented code.

src/lib/ast.rs
```diff
@@ -473,6 +474,46 @@ impl NodeId {
         Ok(())
     }
 
+    /// Attach children to given node
+    pub fn copy_subtree<T: Clone>(&mut self, _to_node: NodeId, position: u16, _arena: &mut Arena<T>) -> ArenaResult {
```
snim2: Is there a reason why so many argument start with an underscore? Usually we only use the underscore if the variable name is used more than once.

snim2:
OK, this is looking really good and most of the comments are very minor! There are a few slightly more involved things we need to sort out before we can merge this:

  * The new method in `ast`, `copy_subtree` needs to have some unit tests associated with it (see the other unit tests at the bottom of the file)

  * For some reason your `rustfmt` produces different results to mine! We need to track that down and fix it. I have version `0.8.6-nightly`, maybe yours is different?

  * The file `new_matcher` should be renamed to something descriptive, like `chawathe98_matcher`. Rather than copying one of the more complete matchers, maybe copy the text of the Null matcher for now (you can find it here: https://github.com/softdevteam/diffract/blob/53fea160b434337664068987443b010b2879c1e1/src/lib/null_matcher.rs )

Once all that is done, we need to bring this branch up to date with master, if you don't have it already it would be a good idea to install meld: http://meldmerge.org/

snim2:
OK, I think you have squashed locally, so go ahead and push the rebased commits, then we can have a go at rebasing against master

k1502606:
This is ready for another review.

src/lib/action.rs
```diff
@@ -59,6 +59,10 @@ pub enum ActionType {
     MOVE,
     /// Update the text or label associated with a node.
     UPDATE,
+    /// Copy takes the node or subtree from tree1 and and copies it into tree2.
```
snim2: Can we say "from tree" and "to tree" here and the comment below?

src/lib/action.rs
```diff
@@ -115,6 +119,21 @@ pub struct Update<T: Clone> {
     label: String,
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
snim2: The `#[derive...` should be below the comment (I realise this is wrong in other parts of the code too!).

src/lib/action.rs
```diff
@@ -115,6 +119,21 @@ pub struct Update<T: Clone> {
...
+/// Copy a subtree from one place to another with an AST.
+pub struct Copy {
+    from_node: NodeId<FromNodeId>,
+    parent: NodeId<FromNodeId>,
+    pos: u16,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
snim2: Here also.

src/lib/ast.rs
```diff
@@ -573,6 +573,50 @@ impl<U: PartialEq + Copy> NodeId<U> {
         Ok(())
     }
 
+    /// Attach children to given node
+    // This would copy the subtree from the to_node
```
snim2: Can we make all these comments `///`?

src/lib/chawathe98_matcher.rs
```diff
@@ -0,0 +1,79 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#![warn(missing_docs)]
+
+/// This matcher performs no operations.
+///
+/// It should ONLY be used for testing the edit script generator, or other parts
```
snim2: Ah, this comment from the null matcher isn't quite true!

src/lib/chawathe98_matcher.rs
```diff
@@ -0,0 +1,79 @@
...
+/// of diffract, and should not be visible to the user -- i.e. it should not be
+/// imported into the `main` module.
+
+use std::fmt::Debug;
+
+use ast::Arena;
+use matchers::{MappingStore, MatchTrees};
+
+#[derive(Debug, Clone, PartialEq)]
+/// The Null matcher needs no configuration.
```
snim2: Let's swap the comment and `derive`

src/lib/chawathe98_matcher.rs
```diff
@@ -0,0 +1,79 @@
...
+pub struct NullConfig {}
```
snim2: Maybe call this `Chawathe98Config` or similar?

snim2:
Ok, just those small comments to go...

src/lib/ast.rs
```diff
@@ -573,6 +573,50 @@ impl<U: PartialEq + Copy> NodeId<U> {
...
+                current_child_node.char_no,
+                current_child_node.token_len,
+            );
+            _new_child_node
+                .make_nth_child_of(_new_node_made, counter, _arena)
+                .unwrap();
+            if !id.is_leaf(_arena) {
+                _new_child_node.copy_subtree(id, counter, _arena).unwrap();
+            }
+            counter = counter + 1;
```
snim2: Maybe replace with `counter += 1`

snim2:
OK, this looks ready to squash into one commit!

PR #59, opened 2017-12-06
Fix issue 57

### LINKED ISSUE #57: Use different types for `NodeId`
Currently, all nodes in ASTs have a unique identifier of type `NodeId`. In both the `from` and `to` ASTs, node ids are numbered from 0. Ideally, it would be better to use separate types for the different tree identifiers (e.g. `FromNodeId` and `ToNodeId`). 

This would help to avoid bugs where we have used a node from the "wrong" tree (several such bugs have already been fixed in this repo!).

/cc @k1502606

snim2:
Fixes #57 

This PR:

  * Introduces separate `FromNodeId` and `ToNodeId` types for `NodeId`, to differentiate *from* and *to* ASTs in the source code. This is intended to avoid bugs where actions are applied to the "wrong" tree.
  * Implements the `From` trait for `NodeId`, `Node` and `Arena` in the `ast` module (only available when testing).
  * Updates Diffract with respect to the lrpar API.
  * Runs `rustfmt` over all affected modules.

It is probably worth reviewing by commit, since the `NodeId` commit is so large.

**NEEDS SQUASHING**

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -41,6 +41,7 @@ use std::collections::VecDeque;
 use std::convert::{TryFrom, TryInto};
 use std::fmt;
 use std::fs::File;
+use std::marker::PhantomData;
 use std::io::Read;
 use std::ops::{Index, IndexMut};
 use std::path::Path;
@@ -53,8 +54,8 @@ use lrtable::{Minimiser, from_yacc};
 
 use hqueue::HeightQueue;
 
-#[derive(Debug)]
 /// Errors raised when parsing a source file.
+#[derive(Debug)]
 pub enum ParseError {
     /// Io error returned from standard library routines.
     Io(String),
@@ -70,8 +71,8 @@ pub enum ParseError {
     SyntaxError,
 }
 
-#[derive(Debug)]
 /// Errors raised by arenas.
+#[derive(Debug)]
 pub enum ArenaError {
     /// Arena is unexpectedly empty.
     EmtpyArena,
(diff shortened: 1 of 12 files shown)


src/lib/ast.rs
```diff
@@ -86,61 +87,98 @@ pub enum ArenaError {
...
+    fn index_mut(&mut self, node: NodeId<U>) -> &mut Node<T, U> {
         &mut self.nodes[node.index]
     }
 }
 
-impl<T: Clone> Arena<T> {
+#[cfg(test)]
+impl<T: Clone> From<Arena<T, ToNodeId>> for Arena<T, FromNodeId> {
+    fn from(arena: Arena<T, ToNodeId>) -> Self {
+        let mut coerced = Vec::<Node<T, FromNodeId>>::new();
```
ltratt: I suggest `with_capacity(arena.size())` instead of `new`.

src/lib/ast.rs
```diff
@@ -86,61 +87,98 @@ pub enum ArenaError {
...
+        for index in 0..arena.size() {
```
ltratt: Maybe we can do this a bit more neatly, something like:

```rust
let coerced = arena.iter()
                   .cloned()
                   .collect::<Vec<Node::<T, FromNodeId>>>()
```

?

src/lib/ast.rs
```diff
@@ -753,10 +854,10 @@ fn read_file(path: &Path) -> Result<String, ParseError> {
 }
 
 /// Parse an individual input file, and return an `Arena` or `ParseError`.
-pub fn parse_file(input_path: &str,
-                  lex_path: &Path,
-                  yacc_path: &Path)
-                  -> Result<Arena<String>, ParseError> {
+pub fn parse_file<T: PartialEq + Copy>(input_path: &str,
+                                lex_path: &Path,
```
ltratt: Not enough indentation?

src/lib/ast.rs
```diff
@@ -681,17 +782,17 @@ impl<'a, T: Clone> Iterator for PreOrderTraversal<'a, T> {
 }
 
 // Turn a grammar, parser and input string into an AST arena.
-fn parse_into_ast(pt: &parser::Node<u16>,
-                  lexer: &Lexer<u16>,
-                  grm: &YaccGrammar,
-                  input: &str)
-                  -> Arena<String> {
+fn parse_into_ast<T: PartialEq + Copy>(pt: &parser::Node<u16>,
+                                lexer: &Lexer<u16>,
```
ltratt: Not enough indentation?

ltratt:
Oops, sorry that I didn't test diffract with the `cfgrammar` change!

snim2:
This is ready for squashing, I thin

snim2:
Squashed.

PR #60, opened 2017-12-11
Use standard terms for type and labels of AST nodes.

### LINKED ISSUE #49: Change nomenclature
The terms from Chawathe et al. `label` and `value` are confusing, and should be replaced by the more commonly used `type` and `value`. This is likely to be a surprisingly large refactoring.

snim2:
**NEEDS SQUASHING**

Fixes #49 

This PR introduces a small API change to rename the `value` of a node as `type` (node `label`s stay the same). 

This has been the cause of much confusion in the code base, and you will notice that there are places where very similar functions have swapped the order types and labels, or where one has been used where the other was expected, and so on.

To avoid further confusion, the unit tests (but not necessarily integration tests) now prefer `&str` to `String` for the type of node types. This avoids confusing between node types and labels and allows the type checker to report any future mistakes.

I've also removed the `hello_java` test from `tests/test_ast.rs` because its expected output was too long to sensibly maintain, and the test did not add anything to the Java nested comment test in the same file.  

GraphViz output still looks OK: [map.pdf](https://github.com/softdevteam/diffract/files/1549724/map.pdf)

### DIFF
--- a/tests/test_ast.rs
+++ b/tests/test_ast.rs
@@ -68,15 +68,15 @@ fn test_empty_calc() {
     compare_ast_dump_to_lrpar_output(
         false,
         "tests/empty.calc",
-        "^~
-  ~
-    WHITESPACE \" \\n\"
-    ~
-  Expr
-    Term
-      Factor
-",
-    );
+        "\"^~\"
+  \"~\"
+    \"WHITESPACE\"  \n
+    \"~\"
+  \"Expr\"
+    \"Term\"
+      \"Factor\"
+"
+  );
 }
 
 #[test]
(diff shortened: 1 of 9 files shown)


tests/test_ast.rs
```diff
@@ -57,6 +57,7 @@ fn compare_ast_dump_to_lrpar_output(is_java: bool, filepath: &str, expected: &st
     };
     let arena: Arena<String, FromNodeId> = parse_file(filepath, lex, yacc).unwrap();
     let arena_pretty_printed = format!("{:?}", arena);
+    println!("\n{}", arena_pretty_printed);
```
ltratt: Unintended print?

snim2: Good spot!

ltratt:
OK, time to squash?

snim2:
Squashed!

PR #64, opened 2018-01-10
Add store ref cells

snim2:
This PR adds interior mutability to mapping stores, so that we can pass them to functions without marking them mutable. I would suggest reviewing each commit separately (note there is a separate formatting commit).

**Needs squashing**

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,1 +1,2 @@
+max_width = 120
 indent_style = "Visual"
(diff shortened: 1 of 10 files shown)


ltratt:
All looks good to me. If you squash, I can merge.

snim2:
Squashed

ltratt:
OK, if you squash this, I think we're ready to merge.

ltratt:
Gah, ignore!

PR #65, opened 2018-01-12
Code improvements suggested by clippy and rustfmt.

snim2:
This should have gone in the last PR. Mea maxima culpa.


ltratt:
Given that the `type`s are only used in one place, does moving them help? It just seems to add a useless level of indirection to me TBH...

snim2:
I don't object to having them as `type`s but I agree, it's not the most useful.

ltratt:
I'd suggest we back that bit back out then -- a smaller diff is no bad thing ;)

snim2:
Needs squashing

ltratt:
Yep, please squash.

snim2:
Squashed

PR #66, opened 2018-01-18
Wrap a RefCell around to_arena in MappingStore.

snim2:
The feature branch `new-edit-script-tests` adds some more tests (mostly integration tests so far) for the edit script generator and contains a (so far partial) fix for the bug in the generator. The code we recently merged into `master`, which wraps a `RefCell` around some elements of the `MappingStore` struct, is especially difficult to merge into the feature branch, and last time I tried I had to give up and abort the merge.

Part of the difficulty is that half-way through the merge we need to add an extra `RefCell` into the `MappingStore` struct (or alternatively mark a number of stores as mutable). Rather than do that in the merge, this PR isolates that change so that we can merge it into master now.

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -298,7 +298,7 @@ impl<T: Clone + Debug + Eq + 'static> Patchify<T> for Move {
                                  f_node.char_no.unwrap(),
                                  f_node.token_len.unwrap()));
         }
-        let t_node = &store.to_arena[store.get_to(&self.from_node).unwrap()];
+        let t_node = &store.to_arena.borrow()[store.get_to(&self.from_node).unwrap()];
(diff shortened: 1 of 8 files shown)


ltratt:
Not that I mind this PR, but I don't understand how this could make merging any easier. Anyway, that doesn't matter too much I suppose.

`src/lib/edit_script.rs.orig` probably shouldn't have been added to the PR?

snim2:
It just means that I don't have to add this extra code during the merge, I can use Meld to merge it in instead.

ltratt:
OK, I still don't get it, but it doesn't matter ;)

If you squash this, I can merge.

snim2:
Thanks, squished.

PR #67, opened 2018-02-05
Induced graph - Added Changes. (Third Stage -- Initial) 

k1502606:
1) Completed the induced graph builder with tests. 
      -> This has several methods which have been implemented from the paper as described. 
2) Completed the pruning method. 
      -> This method will remove edges that violate the rules defined. 
      -> There is also lower bound and upper bound functions which are also used to remove edges.

### DIFF
--- a/src/lib/chawathe98_matcher.rs
+++ b/src/lib/chawathe98_matcher.rs
@@ -46,35 +46,2260 @@
 
 use std::fmt::Debug;
 
-use ast::Arena;
-use matchers::{MappingStore, MatchTrees};
+use ast::{Arena, FromNodeId, NodeId, ToNodeId};
 
-/// The Chawathe98Config matcher needs no configuration.
-#[derive(Debug, Clone, PartialEq)]
-pub struct Chawathe98Config  {}
+// ME -- Hash set
 
-impl Default for Chawathe98Config  {
-    fn default() -> Chawathe98Config  {
-        Chawathe98Config  {}
+use std::collections::HashSet;
+use std::cmp;
+
+/// Not needed by matching algorithms, but useful for debugging.
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub enum EdgeType {
+    /// Check the type of edge update which is update
+    UPDATE,
+    /// Check the type of edge update which is insert
+    INSERT,
+    /// Check the type of edge update which is delete
+    DELETE,
+    /// Check the type of edge update which is move
+    MOVE,
+    /// Check the type of edge update which is copy
+    COPY,
+    /// Check the type of edge update which is glue
+    GLUE,
+    /// Check the type of edge update which is null
+    NULL,
+    /// Check the type of edge update which is Ok
+    OK,
+}
+#[derive(Debug, Clone)]
+/// The intial cost of each edge
+pub struct CostEdge {
+    /// Check the cost of edge update which is update
+    pub cost_update: usize,
+    /// Check the cost of edge update which is insert
+    pub cost_insert: usize,
+    /// Check the cost of edge update which is delete
+    pub cost_delete: usize,
+    /// Check the cost of edge update which is move
+    pub cost_move: usize,
+    /// Check the cost of edge update which is copy
+    pub cost_copy: usize,
+    /// Check the cost of edge update which is glue
+    pub cost_glue: usize,
+    /// Check the cost of edge update which is null
+    pub cost_null: usize,
+    /// Check the cost of edge update which is Ok
+    pub cost_ok: usize,
+}
+
+/// implemenation for edge cost
+
+impl CostEdge {
+    /// Create a new cost edge
+    /// This shows the cost for each edge defined by the user
+    pub fn new(
+        cost_update_1: usize,
+        cost_insert_1: usize,
+        cost_delete_1: usize,
+        cost_move_1: usize,
+        cost_copy_1: usize,
+        cost_glue_1: usize,
+        cost_null_1: usize,
+        cost_ok_1: usize,
+    ) -> CostEdge {
+        CostEdge {
+            cost_update: cost_update_1,
+            cost_insert: cost_insert_1,
+            cost_delete: cost_delete_1,
+            cost_move: cost_move_1,
+            cost_copy: cost_copy_1,
+            cost_glue: cost_glue_1,
+            cost_null: cost_null_1,
+            cost_ok: cost_ok_1,
+        }
     }
 }
 
-impl Chawathe98Config  {
+#[derive(Debug, Clone)]
+/// Edge construct which shows the edge between two nodes and their value
+pub struct Edge {
+    /// from node shows the node id from the From-Tree
+    pub from_node: NodeId<FromNodeId>,
+    /// to node shows the node id from To-Tree
+    pub to_node: NodeId<ToNodeId>,
+    /// The value of the edge
+    pub value: usize,
+    /// The edge type
+    pub edge_type: EdgeType,
+}
+
+/// Implementation for Edges
+impl Edge {
+    /// Create a new edge
+    pub fn new(
+        new_from_node: NodeId<FromNodeId>,
+        new_to_node: NodeId<ToNodeId>,
+        new_value: usize,
+        new_edge_type: EdgeType,
+    ) -> Edge {
+        Edge {
+            from_node: new_from_node,
+            to_node: new_to_node,
+            value: new_value,
+            edge_type: new_edge_type,
+        }
+    }
+}
+
+/// A store of mappings between nodes in different arenas.
+/// Direction is important.
+#[derive(Debug, Clone)]
+pub struct MappingStore2<T: Clone + Debug> {
+    ///List Of edges
+    pub list_edges: Vec<Edge>,
+    /// Source arena (treat as immutable).
+    pub from_arena: Arena<T, FromNodeId>,
+    /// Destination arena (treat as immutable).
+    pub to_arena: Arena<T, ToNodeId>,
+    /// List of nodes to be inserted
+    pub list_node_inserted: Vec<NodeId<ToNodeId>>,
+    /// List of nodes to be deleted
+    pub list_node_deleted: Vec<NodeId<FromNodeId>>,
+    /// Find the Cost of all edges
+    pub all_edge_cost: CostEdge,
+}
+
+impl<T: Clone + Debug + Eq + 'static> MappingStore2<T> {
+    /// Implementation of new in mapping store the second version
+    pub fn new(base: Arena<T, FromNodeId>, diff: Arena<T, ToNodeId>) -> MappingStore2<T> {
+        MappingStore2 {
+            from_arena: base,
+            to_arena: diff,
+            list_edges: Vec::new(),
+            list_node_inserted: Vec::new(),
+            list_node_deleted: Vec::new(),
+            all_edge_cost: CostEdge::new(1, 1, 1, 1, 1, 1, 1, 1),
+        }
+    }
+    /// Implementation of the update of the cost edge.
+    /// This will renew the cost value for each edge type.
+    pub fn update_cost(
+        &mut self,
+        cost_update: usize,
+        cost_insert: usize,
+        cost_delete: usize,
+        cost_move: usize,
+        cost_copy: usize,
+        cost_glue: usize,
+        cost_null: usize,
+        cost_ok: usize,
+    ) {
+        self.all_edge_cost = CostEdge::new(
+            cost_update,
+            cost_insert,
+            cost_delete,
+            cost_move,
+            cost_copy,
+            cost_glue,
+            cost_null,
+            cost_ok,
+        );
+    }
+
+    /// Push New Edge
+    pub fn push(
+        &mut self,
+        new_from_node: NodeId<FromNodeId>,
+        new_to_node: NodeId<ToNodeId>,
+        new_value: usize,
+        new_edge_type: EdgeType,
+    ) {
+        let new_edge = Edge::new(new_from_node, new_to_node, new_value, new_edge_type);
+        self.list_edges.push(new_edge);
+    }
+    /// Remove Edge
+    pub fn remove(&mut self, index: usize) {
+        self.list_edges.remove(index);
+    }
+    /// Check contains the from_node
+    pub fn contains_from(&self, from: NodeId<FromNodeId>) -> bool {
+        for edge in &self.list_edges {
+            if edge.from_node == from {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /// Check if the edges contains
+    ///     1) from node id &
+    ///     2) To node id
+    ///     3) Edge Type
+    pub fn contains_edge(
+        &self,
+        from: usize,
+        to: usize,
+        edge_type: EdgeType,
+        list_edges: &mut Vec<Edge>,
+    ) -> bool {
+        for edge in list_edges {
+            if edge.from_node.id() == from && edge.to_node.id() == to && edge.edge_type == edge_type
+                {
+                    return true;
+                }
+        }
+        return false;
+    }
+
+    /// Check if the edge contains the to_node
+    pub fn contains_to(&self, to: NodeId<ToNodeId>) -> bool {
+        for edge in &self.list_edges {
+            if edge.to_node == to {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /// Print for all the pre order traversal tree2: <Vec<NodeId<ToNodeId>>>
+    pub fn print_preorder_traversal(
+        &self,
+        tree1: Vec<NodeId<FromNodeId>>,
+        tree2: Vec<NodeId<ToNodeId>>,
+    ) -> () {
+        println!();
+        println!("Tree 1");
+        print!("|");
+        for node in tree1 {
+            print!("{:?},{:?}|", node.id(), self.from_arena[node].label);
+        }
+        println!();
+        println!("Tree 2");
+        print!("|");
+        for node in tree2 {
+            print!("{:?},{:?}|", node.id(), self.to_arena[node].label);
+        }
+        println!();
+    }
+    /// Checks if the subtree in node id in From_node is the same same in to_node
+    /// This overall checks the subtrees of two trees and whether they are the same.
+    /// The check is done by checking if the nodes are equal and as well as their descendants.
+    pub fn check_subtree(&self, from_node: NodeId<FromNodeId>, to_node: NodeId<ToNodeId>) -> bool {
+        let mut output = true;
+        let descendants_from_node = from_node
+            .descendants(&self.from_arena)
+            .collect::<Vec<NodeId<FromNodeId>>>();
+        let descendants_to_node = to_node
+            .descendants(&self.to_arena)
+            .collect::<Vec<NodeId<ToNodeId>>>();
+        if eq(&from_node, &self.from_arena, &to_node, &self.to_arena) {
+            if descendants_from_node.len() == descendants_to_node.len() {
+                for i in 0..descendants_to_node.len() {
+                    if !eq(
+                        &descendants_from_node[i],
+                        &self.from_arena,
+                        &descendants_to_node[i],
+                        &self.to_arena,
+                    ) {
+                        // This means the descendants are !perfect by label and expression
+                        output = false;
+                    }
+                }
+            } else {
+                output = false;
+            }
+        } else {
+            output = false;
+        }
+        return output;
+    }
+    /// Check if the subtree is deleted from from_node against to_node.
+    /// This checks whether the the descendants are in the deletion edges.
+    pub fn check_tree_deletion(
+        &self,
+        from_node_ancestor: NodeId<FromNodeId>,
+        delete_tree2: NodeId<ToNodeId>,
+        hash_set_deletion: HashSet<(NodeId<FromNodeId>, NodeId<ToNodeId>)>,
+    ) -> bool {
+        let mut output = true;
+        let descendants_from_node = from_node_ancestor
+            .descendants(&self.from_arena)
+            .collect::<Vec<NodeId<FromNodeId>>>();
+        for descendant in descendants_from_node {
+            if !hash_set_deletion.contains(&(descendant, delete_tree2)) {
+                output = false;
+            }
+        }
+
+        return output;
+    }
+    /// check if subtree 1 exists in tree 2.
+    /// This overall checks from node exists in the same position in tree 2 (To Node Id)
+    /// If it does then true otherwise false.
+    pub fn check_tree_1_exists_in_tree2(&self, from_node_ancestor: NodeId<FromNodeId>) -> bool {
+        let mut output = true;
+        let node_id_to_node = from_node_ancestor.id();
+
+        let to_node = NodeId::<ToNodeId>::new(node_id_to_node);
+        if self.check_subtree(from_node_ancestor, to_node) {
+            //
+            output = true;
+        } else {
+            output = false;
+        }
+
+        return output;
+    }
+    /// Check if the descendants of to_node are in the insertion edges.
+    /// From node would be the insertion node.
+    pub fn check_tree_insertion(
+        &self,
+        insert_tree1: NodeId<FromNodeId>,
+        to_node: NodeId<ToNodeId>,
+        hash_set_insertion: HashSet<(NodeId<FromNodeId>, NodeId<ToNodeId>)>,
+    ) -> bool {
+        let mut output = true;
+        let descendants_from_node = to_node
+            .descendants(&self.to_arena)
+            .collect::<Vec<NodeId<ToNodeId>>>();
+        for descendant in descendants_from_node {
+            if !hash_set_insertion.contains(&(insert_tree1, descendant)) {
+                output = false;
+            }
+        }
+
+        return output;
+    }
+
+    /// Converts vector(To_node_id) to hash set
+    pub fn vector_to_hash_set_to_node_id(
+        &self,
+        tree2: Vec<NodeId<ToNodeId>>,
+    ) -> HashSet<NodeId<ToNodeId>> {
+        let mut hashset = HashSet::<NodeId<ToNodeId>>::new();
+        for node in tree2.clone() {
+            hashset.insert(node);
+        }
+        return hashset;
+    }
+
+    /// Converts vector(From_node_id) to hash set
+    pub fn vector_to_hash_set_from_node_id(
+        &self,
+        tree2: Vec<NodeId<FromNodeId>>,
+    ) -> HashSet<NodeId<FromNodeId>> {
+        let mut hashset = HashSet::<NodeId<FromNodeId>>::new();
+        for node in tree2.clone() {
+            hashset.insert(node);
+        }
+        return hashset;
+    }
+
+    /// Calculate lower bound cost
+    pub fn lower_bound_edge(self, edge1: Edge, edges: Vec<Edge>) -> usize {
+        // Vector to hash map
+        let mut edges_hash_map = HashSet::<(NodeId<FromNodeId>, NodeId<ToNodeId>)>::new();
+
+        for edge in edges {
+            edges_hash_map.insert((edge.from_node, edge.to_node));
+        }
+
+        //
+        // Cost of the lower bound is
+        // Addition of Cost of updating M -> N
+        // Addition of 1/2 * min ()
+
+        // Get the cost of update
+        let cost_update = self.all_edge_cost.cost_update;
+        let cost_move = self.all_edge_cost.cost_move;
+        let m = edge1.from_node.clone();
+        let n = edge1.to_node.clone();
+
+        // Sum of Cost of forced move with m',n where m' is the child of m
+        let m_children = m.children(&self.from_arena)
+            .collect::<Vec<NodeId<FromNodeId>>>();
+        let n_children = n.children(&self.to_arena)
+            .collect::<Vec<NodeId<ToNodeId>>>();
+        let mut cmf_m1_n = 0; // Cost of sum of Cmf(m',n)
+        let mut cmf_m_n1 = 0; // Cost of sum of Cmf(m,n')
+        for m_child in m_children.clone() {
+            let mut calculate_cost = true;
+            /*
+            If n_child does not exist.
+            If M_child & n_child doesn't exist IN Edges.
+                calculate_cost = true
+            IF it exists then there is no calculate cost.
+            */
+            for n_child in n_children.clone() {
+                if edges_hash_map.contains(&(m_child, n_child)) {
+                    calculate_cost = false;
+                }
+            }
+            if calculate_cost {
+                cmf_m1_n += cost_move;
+            }
+        }
+
+        for n_child in n_children.clone() {
+            let mut calculate_cost = true;
+            for m_child in m_children.clone() {
+                if edges_hash_map.contains(&(m_child, n_child)) {
+                    calculate_cost = false;
+                }
+            }
+            if calculate_cost {
+                cmf_m_n1 += cost_move;
+            }
+        }
+
+        let lower_bound_edge = cost_update + 1 / 2 * (cmp::min(cmf_m1_n, cmf_m_n1));
+        return lower_bound_edge;
+    }
+    /// Calculate lower bound cost
+    pub fn upper_bound_edge(self, edge1: Edge, edges: Vec<Edge>) -> usize {
+        // Vector to hash map
+        let mut edges_hash_map = HashSet::<(NodeId<FromNodeId>, NodeId<ToNodeId>)>::new();
+
+        for edge in edges.clone() {
+            edges_hash_map.insert((edge.from_node, edge.to_node));
+        }
+
+        //
+        // Cost of the lower bound is
+        // Addition of Cost of updating M -> N
+        // Addition of 1/2 * min ()
+
+        // Get the cost of update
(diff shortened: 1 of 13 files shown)


src/lib/mod.rs
```diff
@@ -84,3 +84,5 @@ pub mod patch;
 
 /// Algorithms which act on sequences of values.
 pub mod sequence;
+//Testing
```
snim2: Let's replace this comment with a more descriptive one.

src/main.rs
```diff
@@ -390,13 +404,16 @@ fn main() {
...
-                                                                                                                   // This will check the args and compare it with the enum
-                                                                                                                   // There are only 2 enum
-                                                                                                                   // 1) Chawathe96Config
-                                                                                                                   // 2) Chawathe98Config
-                                                                                                                   //
-    fn generator_script_config(input: Option<EditScriptGenerator>) -> Box<edit_script::EditScriptGenerator<String>> {
+    let generator_config: Box<edit_script::EditScriptGenerator<String>> =
+        generator_script_config(args.flag_edit); //  Box::new(edit_script::Chawathe96Config::new());
+    // This will check the args and compare it with the enum
```
snim2: OK, I think we can kill this comment now, because it just describes what the code does.

src/main.rs
```diff
@@ -390,13 +404,16 @@ fn main() {
...
 
-    let generator_config: Box<edit_script::EditScriptGenerator<String>> = generator_script_config(args.flag_edit); //  Box::new(edit_script::Chawathe96Config::new());
-                                                                                                                   // This will check the args and compare it with the enum
-                                                                                                                   // There are only 2 enum
-                                                                                                                   // 1) Chawathe96Config
-                                                                                                                   // 2) Chawathe98Config
-                                                                                                                   //
-    fn generator_script_config(input: Option<EditScriptGenerator>) -> Box<edit_script::EditScriptGenerator<String>> {
+    let generator_config: Box<edit_script::EditScriptGenerator<String>> =
+        generator_script_config(args.flag_edit); //  Box::new(edit_script::Chawathe96Config::new());
```
snim2: Let's kill this commented out code.

src/main.rs
```diff
@@ -390,13 +404,16 @@ fn main() {
     info!("Selecting the Chawathe et al. (1996) edit script generator.");
     // Edit script generator configuration object.
```
snim2: Let's kill this blank line (L406)

src/lib/emitters.rs
```diff
@@ -383,11 +399,138 @@ impl RenderDotfile for MappingStore<String> {
...
+    }
+}
+
+
+
+impl RenderDotfile for MappingStore2<String> {
+    // Because the `dot` crate cannot add arbitrary attributes to nodes and
+    // edges, or create subgraphs and clusters, we render mapping stores to
+    // a dot digraph manually. This produces output similar to the Figure 1
+    // of Falleri et al. (2014).
```
snim2: I'm not sure that "...similar to Figure 1 of Falleri et al. (2014)" is true for your dotfiles? If not let's kill that bit of the comment, or replace it with something more sensible.

src/lib/emitters.rs
```diff
@@ -383,11 +399,138 @@ impl RenderDotfile for MappingStore<String> {
...
+                e1.id()
+            );
+            digraph.push(line);
+        }
+        digraph.push(String::from("\t}\n"));
+        // Mappings between ASTs.
+        let common = "dir=both, arrowsize=.75, arrowhead=odot, arrowtail=odot";
+        let list_edges = self.list_edges.borrow();
+        for edge in list_edges.iter() {
+            //let &(to, ref ty) = val;
```
snim2: Let's kill this commented out code.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
 
-/// The Chawathe98Config matcher needs no configuration.
+use std::cell::RefCell;
+use std::collections::HashSet;
+use std::cmp;
+
+/// Not needed by matching algorithms, but useful for debugging.
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub enum EdgeType {
+    /// Check the type of edge update which is update
```
snim2: These "Check the type" comments in the enum are not so useful. If you want to keep them they need to be rephrased a bit (e.g. "Edge corresponds to an UPDATE operation").

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    GLUE,
+    /// Check the type of edge update which is null
+    NULL,
+    /// Check the type of edge update which is Ok
+    OK,
+}
+#[derive(Debug, Clone)]
+/// The intial cost of each edge
+pub struct CostEdge {
+    /// Check the cost of edge update which is update
```
snim2: The same thing (as above) applies to these comments in the struct.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    pub cost_update: usize,
```
snim2: Because the struct here is called `CostEdge` we know it contains the costs of edges, so I think you can call the struct members `update` rather than `cost_update`, etc.

snim2: Also, I think a `usize` maybe isn't quite what you want here, can a cost really be an arbitrary size? Probably a `u8` or `u16` would be OK?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    pub cost_copy: usize,
+    /// Check the cost of edge update which is glue
+    pub cost_glue: usize,
+    /// Check the cost of edge update which is null
+    pub cost_null: usize,
+    /// Check the cost of edge update which is Ok
+    pub cost_ok: usize,
+}
+
+/// implemenation for edge cost
```
snim2: Let's kill this comment.

src/lib/chawathe98_matcher.rs
ltratt: It's worth thinking carefully about the right size, and using a bigger-rather-than-smaller size if you're not sure. I agree that `usize` is unlikely to be the right choice here, though.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    pub from_node: NodeId<FromNodeId>,
+    /// to node shows the node id from To-Tree
+    pub to_node: NodeId<ToNodeId>,
+    /// The value of the edge
+    pub value: usize,
+    /// The edge type
+    pub edge_type: EdgeType,
+}
+
+/// Implementation for Edges
```
snim2: Let's kill this comment also

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+            value: new_value,
+            edge_type: new_edge_type,
+        }
+    }
+}
+
+/// A store of mappings between nodes in different arenas.
+/// Direction is important.
+#[derive(Debug, Clone)]
+pub struct MappingStore2<T: Clone + Debug> {
```
snim2: Can we call this something more descriptive than `MappingStore2`? Maybe `MappingStoreGraph` or `MappingStoreInducedGraph` or similar?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    ///List Of edges
```
snim2: Space between `/` and `L`

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    /// Source arena (treat as immutable).
+    pub from_arena: RefCell<Arena<T, FromNodeId>>,
+    /// Destination arena (treat as immutable).
+    pub to_arena: RefCell<Arena<T, ToNodeId>>,
+    /// List of nodes to be inserted
+    pub list_node_inserted: RefCell<Vec<NodeId<ToNodeId>>>,
+    /// List of nodes to be deleted
+    pub list_node_deleted: RefCell<Vec<NodeId<FromNodeId>>>,
+    /// Find the Cost of all edges
+    pub all_edge_cost: RefCell<CostEdge>,
```
snim2: I'm not sure you need the `RefCell` here. The reason we have it in the other mapping store, is that sometimes we need to hold two references to one of the structures. For example, we might want to iterate over one of the arenas, and also mutate the "current" node that we're looking at. I would guess you don't need anything that complicated for the `CostEdge` struct, but I could be wrong!

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+impl<T: Clone + Debug + Eq + 'static> MappingStore2<T> {
+    /// Implementation of new in mapping store the second version
+    pub fn new(base: Arena<T, FromNodeId>, diff: Arena<T, ToNodeId>) -> MappingStore2<T> {
+        MappingStore2 {
+            from_arena: RefCell::new(base),
+            to_arena: RefCell::new(diff),
+            list_edges: RefCell::new(Vec::new()),
+            list_node_inserted: RefCell::new(Vec::new()),
+            list_node_deleted: RefCell::new(Vec::new()),
+            all_edge_cost: RefCell::new(CostEdge::new(1, 1, 1, 1, 1, 1, 1, 1)),
```
snim2: This is fine, but because it's a bit of a temporary hack, let's put a comment in to say "Currently, we set all the edge costs to be 1, but in future they will be configurable."

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        for edge in list_edges.iter() {
+            if edge.to_node == to {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /// Print for all the pre order traversal tree2: <Vec<NodeId<ToNodeId>>>
+    pub fn print_preorder_traversal(
```
snim2: There are a few places where you have different "print" type functions. I think for now let's leave them, but in general in Rust we would implement the `fmt::Display` or `fmt::Debug` trait on a struct, and then it would be possible to just say `println!("{:?}", my_struct_instance);`. It might be worth raising an issue for this and coming back to it another day, but I think this PR is big enough that we don't want to make this particular change here.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        for node in tree2 {
+            print!("{:?},{:?}|", node.id(), self.to_arena.borrow()[node].label);
+        }
+        println!();
+    }
+    /// Checks if the subtree in node id in From_node is the same same in to_node
+    /// This overall checks the subtrees of two trees and whether they are the same.
+    /// The check is done by checking if the nodes are equal and as well as their descendants.
+    pub fn check_subtree(&self, from_node: NodeId<FromNodeId>, to_node: NodeId<ToNodeId>) -> bool {
+        let mut output = true;
```
snim2: I'm not sure you need this variable?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        let descendants_from_node = from_node
+            .descendants(&self.from_arena.borrow())
+            .collect::<Vec<NodeId<FromNodeId>>>();
+        let descendants_to_node = to_node
+            .descendants(&self.to_arena.borrow())
+            .collect::<Vec<NodeId<ToNodeId>>>();
+        if eq(&from_node, &self.from_arena.borrow(), &to_node, &self.to_arena.borrow()) {
```
snim2: I'm not sure that I've understood this comparison correctly. Does it compare two `ast::Node` types? If so, I doubt that is what you want to do. The `Node` struct looks like this (I've removed the comments):

```rust
pub struct Node<T: Clone, U: PartialEq + Copy> {
    parent: Option<NodeId<U>>,
    previous_sibling: Option<NodeId<U>>,
    next_sibling: Option<NodeId<U>>,
    first_child: Option<NodeId<U>>,
    last_child: Option<NodeId<U>>,
    index: Option<NodeId<U>>,
    pub label: String,
    pub ty: T,
    pub char_no: Option<usize>,
    pub col_no: Option<usize>,
    pub line_no: Option<usize>,
    pub token_len: Option<usize>,
}
```

So, much of what you are comparing is `NodeId` ids, but subtrees can be "the same" in the two ASTs even if their node ids are different. I think what you are looking for here is something like the `is_isomorphic()` function in the `matchers.rs` file? Either way, this is a function that will need careful testing!

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        let mut output = true;
+        let descendants_from_node = from_node_ancestor
+            .descendants(&self.from_arena.borrow())
+            .collect::<Vec<NodeId<FromNodeId>>>();
+        for descendant in descendants_from_node {
+            if !hash_set_deletion.contains(&(descendant, delete_tree2)) {
+                output = false;
+            }
+        }
+
```
snim2: Extra blank line.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    /// check if subtree 1 exists in tree 2.
+    /// This overall checks from node exists in the same position in tree 2 (To Node Id)
+    /// If it does then true otherwise false.
+    pub fn check_tree_1_exists_in_tree2(&self, from_node_ancestor: NodeId<FromNodeId>) -> bool {
+        let mut output = true;
+        let node_id_to_node = from_node_ancestor.id();
+
+        let to_node = NodeId::<ToNodeId>::new(node_id_to_node);
+        if self.check_subtree(from_node_ancestor, to_node) {
+            //
```
snim2: Let's kill this empty comment!

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    /// Calculate lower bound cost
+    pub fn lower_bound_edge(self, edge1: Edge, edges: Vec<Edge>) -> usize {
+        // Vector to hash map
+        let mut edges_hash_map = HashSet::<(NodeId<FromNodeId>, NodeId<ToNodeId>)>::new();
+
+        for edge in edges {
+            edges_hash_map.insert((edge.from_node, edge.to_node));
+        }
+
+        //
```
snim2: Empty comment

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        let mut cmf_m_n1 = 0; // Cost of sum of Cmf(m,n')
+        for m_child in m_children.clone() {
+            let mut calculate_cost = true;
+            /*
+            If n_child does not exist.
+            If M_child & n_child doesn't exist IN Edges.
+                calculate_cost = true
+            IF it exists then there is no calculate cost.
+            */
+            for n_child in n_children.clone() {
```
snim2: Do you really need these clones?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    /// Calculate lower bound cost
+    pub fn upper_bound_edge(self, edge1: Edge, edges: Vec<Edge>) -> usize {
+        // Vector to hash map
+        let mut edges_hash_map = HashSet::<(NodeId<FromNodeId>, NodeId<ToNodeId>)>::new();
+
+        for edge in edges.clone() {
+            edges_hash_map.insert((edge.from_node, edge.to_node));
+        }
+
+        //
```
snim2: Empty comment

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+
+        // Sum of Cost of forced move with m',n where m' is the child of m
+        let m_children = m.children(&self.from_arena.borrow())
+            .collect::<Vec<NodeId<FromNodeId>>>();
+        let n_children = n.children(&self.to_arena.borrow())
+            .collect::<Vec<NodeId<ToNodeId>>>();
+        let mut cm_m1_n = 0; // Cost of sum of Cmf(m',n)
+        let mut cm_m_n1 = 0; // Cost of sum of Cmf(m,n')
+        for m_child in m_children.clone() {
+            //
```
snim2: Same, and are the `clone()`s needed?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+                }
+            }
+            em1 = em1 - 1;
+
+            for n_child in n_children.clone() {
+                if edges_hash_map.contains(&(m_child, n_child)) {
+                    calculate_cost = true;
+                }
+            }
+            //
```
snim2: Empty comment

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+            if calculate_cost {
+                cm_m1_n += cost_move;
+            }
+            //
```
snim2: Empty comment

snim2: And a few below!

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+                else if list_m.len() == 1 && list_n.len() > 1 {
+                    // //// CASE 2.1 //// //
+                    // Figure 16, appendix D
+                    if list_m[0].id() == m.id() && m.id() != self.from_arena.borrow().size() - 1 {
+                        // Partition
+                        for n1 in &list_n {
+                            for n2 in &list_n {
+                                // Check if n1.parent == n2.parent are the same
+                                if n1.id() != n2.id() {
+                                    //                                    let n1_parent = self.from_arena[m].parent().unwrap();
```
snim2: Let's remove any commented out code

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+                                    //                                    let n2_parent = self.to_arena[n].parent().unwrap();
+                                }
+                            }
+                        }
+                    }
+                        // //// CASE 2.2 //// //
+                        // Figure 17, appendix D
+                        else if list_m[0].id() == self.from_arena.borrow().size() - 1 {
+
```
snim2: It looks like this `if` statement and the one above don't do anything? i.e. there's no code inside the braces?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+            index += 1;
+            println!("--");
+        }
+    }
+}
+
+/// The Chawathe98Config2 matcher needs no configuration.
 #[derive(Debug, Clone, PartialEq)]
-pub struct Chawathe98Config  {}
+pub struct Chawathe98Config2 {}
```
snim2: It's unfortunate that we end up having two of these configs, but can we call them something other than `Config2`?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        diff: Arena<T, ToNodeId>,
+    ) -> MappingStore2<T> {
+        let mut store = MappingStore2::new(base.clone(), diff.clone());
+        if store.from_arena.borrow().is_empty() || store.to_arena.borrow().is_empty() {
+            return store;
+        }
+
+        let last_node_from_node = NodeId::<FromNodeId>::new(base.size() - 1);
+        let last_node_to_node = NodeId::<ToNodeId>::new(diff.clone().size() - 1);
+        //let last_node_from_node = base[size_fromNode];
```
snim2: In general, we don't want to merge commented out code into master, so let's kill these lines, and the ones below

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+            }
+        }
+
+        store
+    }
+
     /// Describe this matcher for the user.
     fn describe(&self) -> String {
-        let desc = "This matcher performs no matching operations.";
+        let desc = "This matcher performs matching operation from the paper Chawathe98.";
```
snim2: `Chawathe et al. (1998)`

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+            // Get All The edges for E3 ---> where (m1,n)
+            // Where the to_node_id is the same but to from_id is different
+            let mut upper_bound_e2_and_e3 = 0;
+
+            // E2 = (m,n1) &&   E3 = (m1,n)
+
+            upper_bound_e2_and_e3 = input
+                .clone()
+                .upper_bound_edge(edge.clone(), all_edges.clone());
+            //
```
snim2: Let's kill these empty comments

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+                edges_after_pruning_update.push(e1_update.clone());
+            }
+        }
+
+        edges_still_left = edges_after_pruning_update;
+
+        // The Root should not move just update.
+        // If it's root it should not have move, then it should have insert as well
+        //          Which doesn't make any sense
+        // It could have glue, copy
```
snim2: I'm not sure I follow this comment. What happens if the two root nodes have different types and different labels?

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+    /// CLI option.
+    fn describe(&self) -> String;
+
+    /// This shows the trait for the pruning for the induced edges.
+    /// The result should logically reduce the number of edges produced in induced edges.
+    fn prune_edges(&self, input: MappingStore2<T>) -> (Vec<Edge>, Vec<Edge>);
+}
+
+#[cfg(test)]
+
```
snim2: Let's kill this empty line

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        Questions.
+                 1) Do we need to consider where when there multiple (m,n) edge ?
+        */
+
+        return (edges_still_left, edges_to_be_pruned);
+    }
+}
+
+/// Creates a simple tree for tests which is FromNodeId
+pub fn create_arena() -> Arena<String, FromNodeId> {
```
snim2: I think this code is used in your tests and not in the core algorithms, so it needs to be moved into the test module below that starts: `mod tests`.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+}
+
+/// Test that two nodes have the same label.
+fn eq_label<T: Clone + Debug + Eq>(
+    n1: &NodeId<FromNodeId>,
+    arena1: &Arena<T, FromNodeId>,
+    n2: &NodeId<ToNodeId>,
+    arena2: &Arena<T, ToNodeId>,
+) -> bool {
+    arena1[*n1].ty == arena2[*n2].ty
```
snim2: This tests the types of the nodes, not their labels! So either the code here is wrong or the function name and comment!

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+}
+
+/// Match two trees and return a store of mappings between them.
+///
+/// This trait should usually be implemented on configuration objects that
+/// define thresholds and weights for a given algorithm.
+pub trait MatchTrees2<T: Clone + Debug> {
```
snim2: OK, let's rename this somehow and get rid of the `2`.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+}
+
+#[cfg(test)]
+
+mod tests {
+    use super::*;
+
+    #[test]
+    /// Test for perfect Matching
+    /// TEST 1
```
snim2: Let's remove `TEST 1` at least; and put the other line of comment above the `#[test]`.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        let matcher = matching_config.match_trees(tree1.clone(), tree2.clone());
+        let mut checker_edges_induced = matcher.list_edges.borrow().clone();
+
+        //    ASSERT START
+        assert!(matcher.contains_edge(0, 0, EdgeType::OK, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(1, 1, EdgeType::OK, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(2, 2, EdgeType::OK, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(3, 3, EdgeType::OK, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(4, 4, EdgeType::OK, &mut checker_edges_induced,));
+        //      Assert END
```
snim2: Let's remove these comments and commented out code.

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        // Visual Test
+        //        let fileLocation = "./test_AS_1.dot";
+
+        //        write_dotfile_to_disk(fileLocation, &matcher);
+    }
+
+    #[test]
+    // Tree 1 has more nodes and tree 2 stayst he same as create arena
```
snim2: Same comments as the test above...

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        let n3 = arena.new_node(
+            String::from("Expr"),
+            String::from("d"),
+            None,
+            None,
+            None,
+            None,
+        );
+        n3.make_child_of(n2, &mut arena).unwrap();
+        /*
```
snim2: These visual comments showing the shape of the AST are very nice!

src/lib/chawathe98_matcher.rs
```diff
@@ -46,35 +46,2262 @@
...
+        assert!(matcher.contains_edge(10, 7, EdgeType::MOVE, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(10, 10, EdgeType::UPDATE, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(11, 8, EdgeType::MOVE, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(12, 9, EdgeType::MOVE, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(13, 10, EdgeType::MOVE, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(9, 6, EdgeType::GLUE, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(4, 1, EdgeType::GLUE, &mut checker_edges_induced,));
+        assert!(matcher.contains_edge(1, 3, EdgeType::GLUE, &mut checker_edges_induced,));
+        // Visual Check
+        /
```
snim2: Commented out code...

snim2:
OK, I think that concludes looking at all the code. Most of the comments are very minor, the only couple of things that look "wrong" as such, are the function that says it compares node labels but actually compares node types, and the function that checks whether subtrees are "equal", which I think probably should be something like the isomorphism test in `matchers.rs`. Otherwise it's just small changes. 

A couple of general things:

  * Most of us comment out code while we're debugging, and it doesn't matter if some of that slips into a PR (and it seems to happen reasonably often to all of us), but we do need to remove these comments before they get merged with master.
  * Naming things and commenting are hard! In general, you want to tell the reader about the things that aren't obvious from looking at the code. For example, there's a function called something like `detach` in the AST module. Probably a programmer will figure out from the name that detatching a node means removing it from its parent, but it isn't obvious what should happen to the children of that node, or whether the node stays in the `Arena` if it no longer has any links to other nodes, so those things need to be made clear in the comments. 

There are quite a few formatting issues in this PR, and you could deal with them all at once by using `rustfmt`. I'm not sure that your copy of `rustfmt` is behaving like mine does though, and really we want all the code in the repo to be formatted in the same way. There is a file called `rustfmt.toml` which is supposed to control some options, and that should help. My rust is like this:

```sh
$ rustc -V
rustc 1.25.0-nightly (61452e506 2018-01-09)
$ rustfmt -V
0.3.4-nightly ( )
$ 
```

so you may need to update Rust, and install `rustfmt-nightly` if you don't have it already?

src/lib/chawathe98_matcher.rs
```diff
@@ -288,54 +266,56 @@ impl<T: Clone + Debug + Eq + 'static> MappingStore2<T> {
...
     pub fn check_subtree(&self, from_node: NodeId<FromNodeId>, to_node: NodeId<ToNodeId>) -> bool {
-        let mut output = true;
-        let descendants_from_node = from_node
-            .descendants(&self.from_arena.borrow())
-            .collect::<Vec<NodeId<FromNodeId>>>();
-        let descendants_to_node = to_node
-            .descendants(&self.to_arena.borrow())
-            .collect::<Vec<NodeId<ToNodeId>>>();
-        if eq(&from_node, &self.from_arena.borrow(), &to_node, &self.to_arena.borrow()) {
+        //        let mut output = true;
```
snim2: Let's kill this commented code

src/lib/chawathe98_matcher.rs
```diff
@@ -288,54 +266,56 @@ impl<T: Clone + Debug + Eq + 'static> MappingStore2<T> {
...
-                    ) {
+                    if !eq(&descendants_from_node[i],
+                           &self.from_arena.borrow(),
+                           &descendants_to_node[i],
+                           &self.to_arena.borrow())
+                    {
                         // This means the descendants are !perfect by label and expression
-                        output = false;
+                        return false;
+                        //output = false;
```
snim2: And this line

src/lib/chawathe98_matcher.rs
```diff
@@ -1085,139 +980,31 @@ impl<T: Clone + Debug + Eq + 'static> MatchTrees2<T> for Chawathe98Config2 {
...
 
-/// Test that two nodes have the same label.
-fn eq_label<T: Clone + Debug + Eq>(
-    n1: &NodeId<FromNodeId>,
-    arena1: &Arena<T, FromNodeId>,
-    n2: &NodeId<ToNodeId>,
-    arena2: &Arena<T, ToNodeId>,
-) -> bool {
+/// Test that two nodes have the same types.
+fn eq_label<T: Clone + Debug + Eq>(n1: &NodeId<FromNodeId>,
```
snim2: Maybe we should call this function `eq_type`?!

src/lib/chawathe98_matcher.rs
```diff
@@ -1085,139 +980,31 @@ impl<T: Clone + Debug + Eq + 'static> MatchTrees2<T> for Chawathe98Config2 {
...
+                                   -> bool {
     arena1[*n1].ty == arena2[*n2].ty
 }
 
 /// Match two trees and return a store of mappings between them.
 ///
 /// This trait should usually be implemented on configuration objects that
 /// define thresholds and weights for a given algorithm.
-pub trait MatchTrees2<T: Clone + Debug> {
+pub trait MatchingTreesScriptor<T: Clone + Debug> {
```
snim2: Do you mean `MatchTreesDescriptor`?

src/lib/chawathe98_matcher.rs
snim2: I think this comment hasn't been actioned?

src/lib/chawathe98_matcher.rs
```diff
@@ -1227,319 +1014,332 @@ pub trait MatchTrees2<T: Clone + Debug> {
...
-        );
+        let n6 = arena.new_node(String::from("INT"),
+                                String::from("4"),
+                                None,
+                                None,
+                                None,
+                                None);
         n6.make_child_of(n4, &mut arena).unwrap();
 
         //
```
snim2: Empty comment

src/lib/chawathe98_matcher.rs
```diff
@@ -167,54 +159,49 @@ pub struct MappingStore2<T: Clone + Debug> {
...
-            list_node_deleted: RefCell::new(Vec::new()),
-            all_edge_cost: RefCell::new(CostEdge::new(1, 1, 1, 1, 1, 1, 1, 1)),
-        }
+    pub fn new(base: Arena<T, FromNodeId>, diff: Arena<T, ToNodeId>) -> MappingStoreGraph<T> {
+        MappingStoreGraph { from_arena: RefCell::new(base),
+                            to_arena: RefCell::new(diff),
+                            list_edges: RefCell::new(Vec::new()),
+                            list_node_inserted: RefCell::new(Vec::new()),
+                            list_node_deleted: RefCell::new(Vec::new()),
+                            /* The cost could change in the future but,
```
snim2: This comment is rather oddly laid out!

snim2:
OK, this is nearly there I think! There are some tests to fix and I'm not sure about `rustfmt` yet, but I think we're close to being able to merge now.

k1502606:
Hi, if you get time would you please review the changes. Thanks.

snim2:
Ok, great! Time to squash this down.

src/lib/chawathe98_matcher.rs
```diff
@@ -170,9 +170,8 @@ impl<T: Clone + Debug + Eq + 'static> MappingStoreGraph<T> {
                             list_edges: RefCell::new(Vec::new()),
                             list_node_inserted: RefCell::new(Vec::new()),
                             list_node_deleted: RefCell::new(Vec::new()),
-                            /* The cost could change in the future but,
-                ->  we can make the default cost for every edge cost the value of 1.
-            */
+            //The cost could change in the future but,
```
snim2: Can we add a space after the `//`?

snim2:
OK, let's squash!

PR #76, opened 2018-02-27
Fix typo in test dealing with non-existent lexer file

snim2:
This bug was not caught previously, because the intended file not found failure (i.e. the missing lexer file) was triggered before the missing input file caused a panic.


ltratt:
Can you rebase?

PR #77, opened 2018-02-27
Action module improvements

snim2:
This PR:

  * Adds a new `size()` method, 
  * Implements the `Debug` trait, and
  * Implements the `PartialEq` 

for `EditScript` objects. The last of these is complicated, because we store edit scripts as a vector of boxed traits. These need to be downcast before calling `eq()` and we need to avoid unbounded recursion in the `eq()` method.

See also [this SO question](https://stackoverflow.com/questions/48952627/implementing-partialeq-on-a-boxed-trait).

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -39,7 +39,7 @@
 
 use std::boxed::Box;
 use std::clone::Clone;
-use std::fmt::Debug;
+use std::fmt;
 
 use ast::{Arena, ArenaError, ArenaResult, FromNodeId, NodeId};
 use emitters::RenderJson;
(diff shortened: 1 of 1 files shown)


snim2:
Needs squashing.

ltratt:
Why do we need the `as_any` function?

snim2:
For type up/down-casting, so that we can call `eq()`

ltratt:
Ah, OK, I see. I guess this begs the question of whether the trait approach is the right one here. What might the alternatives be?

snim2:
As far as I can tell, this is the idiomatic way to have a list of polymorphic objects. Possibly we could replace the action structs with enums, or in some cases it is apparently possible to replace the `Box` with a reference. This probably isn't a job for this PR though!

snim2:
See Issue #78

PR #81, opened 2018-02-28
Improve edit script tests

### LINKED ISSUE #68: Glob files and apply macro for integration tests
Currently in integration tests such as `tests/test_edit_script.rs` we have a boilerplate function which applies the tests, then a number of functions which apply the boilerplate to the input files.

It would be nicer to glob the test files in the `tests/` directory (currently `*.calc`, `*.txt` and `*.java`) and apply the tests with a macro, as in https://stackoverflow.com/questions/34662713/how-can-i-create-parameterized-tests-in-rust/34666891#34666891

This needs a little thought, we would need to automatically select the correct lexer and parser, based on the filename, as we do in `main` and `ast`. This code might be better factored out since we already use it in two places in the code base.

Globbing requires an external crate: https://crates.io/crates/glob

snim2:
This PR:

  * Adds some simple unit tests to `edit_script`
  * Adds many integration tests to `test_edit_script`
  * Factors out code related to parsing (including code for finding lex and yacc files) which is common to several tests and other parts of the code base
  * Adds a slow-running integration test which tests every pair of example files in the `tests` directory (test is ignored by default), and thus fixes #68 

Several unit and integration tests are currently failing and ignored by default, these will be fixed in a later PR (many have already been fixed on the `new_edit_script_tests` branch).


src/lib/null_matcher.rs
```diff
@@ -0,0 +1,79 @@
...
+use std::fmt::Debug;
+
+use ast::{Arena, FromNodeId, ToNodeId};
+use matchers::{MappingStore, MatchTrees};
+
+#[derive(Debug, Clone, PartialEq)]
+/// The Null matcher needs no configuration.
+pub struct NullConfig {}
+
+impl Default for NullConfig {
```
ltratt: Do we ever use `Default` on this? If not, let's remove it. [I fear the "default" warning from Clippy causes much more harm than good!]

snim2:
OK, will need squashing.

ltratt:
I'm not convinced that the "run millions of tests in one test" thing is a great idea. I imagine it makes Cargo complain that the test has overrun? I wonder if we need the idea of a "small" and a "large" test suite?

snim2:
It complains that the test runs >60secs but carries on running it. As the comment says, it's not run by default, you have to run `cargo test --no-fail-fast  -- --ignored --nocapture`. It would be nicer to have a macro that creates one "unit" test per pair of files (well two, one for each matcher) but I couldn't figure out how to make that work.

ltratt:
I wonder how other projects with long tests deal with this?

snim2:
I couldn't figure that out, apart from using `#[ignore]`. I did find this about creating tests with macros: https://stackoverflow.com/questions/34662713/how-can-i-create-parameterized-tests-in-rust/34666891#34666891

ltratt:
Yes, although you'd need to manually enumerate the filenames I think (i.e. I don't believe you can write a macro which interacts with the filesystem). If that isn't a problem, it might be the neater solution.

snim2:
Ok, I have re-implemented the auto-generated tests based on [this SO answer](https://stackoverflow.com/a/49056967/342327). It isn't quite as neat a solution as I'd hoped but it works. 

This is ready for re-review (but needs squashing).

build.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Directory containing example files for testing and to which the generated
+/// tests should be written.
+const DIR: &str = "tests/";
+
+lazy_static! {
+    /// Glob all example files in the `tests/` directory.
+    /// Filter out files which have been seeded with deliberate lexical or
+    /// syntactic errors.
+    static ref TEST_FILES: Vec<String> = glob(&format!("{}*.calc", DIR))
```
ltratt: What's the point of making this static? Since this program is run once, this just seems like extra complexity (and an extra dependency with `lazy_static`)?

build.rs
```diff
@@ -0,0 +1,133 @@
...
+
+use diffract::myers_matcher::MyersConfig;
+use diffract::null_matcher::NullConfig;
+
+mod common;
+use common::check_files;
+
+    "#
+    ).unwrap();
+    let mut duplicates: Vec<&str> = vec![]; // Avoid duplicate unit tests.
```
ltratt: Why would we have duplicate tests? If we find one, shouldn't we error?

tests/common.rs
```diff
@@ -0,0 +1,86 @@
...
+extern crate diffract;
+
+use diffract::edit_script::{Chawathe96Config, EditScriptGenerator};
+use diffract::matchers::MatchTrees;
+use diffract::parser::{get_lexer, get_parser, parse_file};
+
+/// Check that the Chawathe 1996 edit script generator correctly generates an
+/// edit script for two files and a given matcher.
+pub fn check_files(path1: &str, path2: &str, matcher: Box<MatchTrees<String>>) {
+    // Lex and parse the input files.
```
ltratt: Not sure this comment is necessary.

tests/common.rs
```diff
@@ -0,0 +1,86 @@
...
+    let ast_from = parse_file(path1, &get_lexer(path1), &get_parser(path1)).unwrap();
+    let ast_to = parse_file(path2, &get_lexer(path2), &get_parser(path2)).unwrap();
+    // Generate mappings between ASTs.
+    let store = matcher.match_trees(ast_from.clone(), ast_to.clone());
+    // Generate an edit script.
+    let gen_config: Box<EditScriptGenerator<String>> = Box::new(Chawathe96Config::new());
+    let edit_script_wrapped = gen_config.generate_script(&store);
+    assert!(edit_script_wrapped.is_ok(),
+            "Edit script generator failed to complete.");
+    // Get root nodes after the edit script has run (note that the root nodes
```
ltratt: Are these the root nodes or root node IDs? I assume the latter, otherwise I'm not sure how anything could be mutated.

ltratt:
OK, a few comments, although nothing tremendously deep.

build.rs
snim2: I think this was left over from a time when the test names were less sophisticated, so files like `Hello.txt` and `Hello.java` would produce four tests called `test_globbed_hello_hello`!

tests/common.rs
snim2: They are root node ids. To avoid all sorts of fun with the borrow checker, many methods in `ast.rs` are methods on IDs, which pass the arena (the AST) to the method.

snim2:
The fist Travis test has passed; is this ready for squashing?

build.rs
```diff
@@ -84,6 +60,20 @@ lazy_static! {
 ///
 /// Run with: `cargo test --no-fail-fast  -- --ignored --nocapture`
 fn main() {
+    // Glob all example files in the `tests/` directory.
+    // Filter out files which have been seeded with deliberate lexical or
+    // syntactic errors.
+    let test_files: Vec<String> =
+        glob(&format!("{}*.calc", DIR)).expect("Failed to read glob pattern")
+                                       .into_iter()
+                                       .chain(glob(&format!("{}*.java", DIR)).expect("Failed to read glob pattern")
```
ltratt: I think these lines are a bit too long (and I quite like long lines :) ).

ltratt:
Just that one comment, then I think we'll be done.

snim2:
ok?

ltratt:
Yep, please squash.

snim2:
Squashed

snim2:
Hmm, Travis thinks that final test has passed and GH does not...

ltratt:
I suggest kicking Travis to rerun things and seeing if that solves the problem.

snim2:
OK, kicked...

snim2:
Bingo!

PR #83, opened 2018-03-08
Edit script fixes

snim2:
This PR contains a number of fixes for the edit script generator. It is probably easier to review commit-by-commit, and the commit messages contain details of each fix.

The only commit that needs some extra information is 5bfe448a1f18d3f41a7a2dda0ea1e569bf0bebc0 which "fixes" something that looked like a bug, but was actually something that makes no difference to the algorithms, but produced very confusing debug output. Here is an example from master branch: 

[master_edit.pdf](https://github.com/softdevteam/diffract/files/1794497/master_edit.pdf)
 
In this example, we have diff'd two files of different file types, meaning that in the original ASTs the two trees will have had different roots (with different types and labels). The Chawathe et al. 96 algorithm assumes that trees have the same root, and so while the edit script generators is running, we add in temporary roots to both trees, and remove them once the edit script is complete. The mistake here was to leave the new root of the tree with a parent, which is the temporary root node. You can see in the PDF that the temporary root is coloured grey, meaning that it has been deleted from the AST, but because it is still referred to it appears in the tree.

Here is the same example in this branch:

[feature_edit.pdf](https://github.com/softdevteam/diffract/files/1794508/feature_edit.pdf)

The ignored tests (the long run of tests for every pair of example files in the `tests/` directory) also pass.

### DIFF
--- a/src/lib/edit_script.rs
+++ b/src/lib/edit_script.rs
@@ -153,13 +153,13 @@ impl Chawathe96Config {
             from_in_order.insert(a);
             to_in_order.insert(b);
         }
-        // 6. For each a in s1, b in s2 such that (a, b) in store (M) but
+        // 6. For each a in s1, b in s2 such that (a, b) in store (M') but
         // (a, b) not in lcs: (i) let k = find_pos(b); (ii) append MOV(a, w, k)
         // to the script and apply MOV(a, w, k) to T_1; (iii) Mark a, b as "in
-        // order".
+        // order". Note, the paper says M (for the store), but should say M'.
         for a in &s1 {
             for b in &s2 {
-                if store.is_mapped_by_matcher(a, b) && !lcs.contains(&(*a, *b)) {
+                if store.is_mapped(a, b) && !lcs.contains(&(*a, *b)) {
                     let k = self.find_pos(store, *b, from_in_order, to_in_order);
                     let mut mov = Move::new(*a, w, k);
                     debug!("Edit script align_children: MOV {:?} {} Parent: {:?} {} Child: {}",
--- a/src/lib/edit_script.rs
+++ b/src/lib/edit_script.rs
@@ -196,7 +196,8 @@ impl Chawathe96Config {
         let x_pos = x.get_child_position(&store.to_arena.borrow()).unwrap();
         // 3. Find v in T_2, where v is the rightmost sibling of x that is to
         // the left of x and is marked "in order".
-        let v = x.children(&store.to_arena.borrow()).take(x_pos)
+        let v = siblings.iter()
+                 .take(x_pos)
                  .filter(|c| to_in_order.contains(c))
                  .last();
         if v.is_none() {
--- a/src/lib/edit_script.rs
+++ b/src/lib/edit_script.rs
@@ -45,7 +45,11 @@ use action::{ApplyAction, Delete, EditScript, Insert, Move, Update};
 use ast::{ArenaError, FromNodeId, NodeId, ToNodeId};
 use matchers::{EditScriptResult, MappingStore, MappingType};
 
-const TMP_ROOT: &str = "_____DIFFRACT_TMP_ROOT";
+/// Label to use when introducing temporary root nodes in an AST.
+///
+/// Temporary root nodes should be removed before the edit script generator
+/// completes. This constant is exposed for use in integration tests.
+pub const TMP_ROOT: &str = "_____DIFFRACT_TMP_ROOT";
 
 /// Given a matching between two ASTs, generate a complete edit script.
 ///
--- a/src/lib/edit_script.rs
+++ b/src/lib/edit_script.rs
@@ -283,6 +283,9 @@ impl<T: Clone + Debug + Default + Eq + 'static> EditScriptGenerator<T> for Chawa
         let mut script: EditScript<T> = EditScript::new();
         let mut from_in_order: HashSet<NodeId<FromNodeId>> = HashSet::new();
         let mut to_in_order: HashSet<NodeId<ToNodeId>> = HashSet::new();
+        if store.from_arena.borrow().is_empty() && store.to_arena.borrow().is_empty() {
+            return Ok(script);
+        }
         // Combined update, insert, align and move phases.
         // 2. Visit the nodes of T_2 in breadth-first order.
         assert!(store.from_arena.borrow().root().is_some(),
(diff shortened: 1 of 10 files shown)


src/lib/edit_script.rs
```diff
@@ -178,7 +185,8 @@ impl Chawathe96Config {
                                                  to_in_order: &HashSet<NodeId<ToNodeId>>)
                                                  -> u16 {
         // 1. Let y=p(x) in T_2 and let w be the partner of x in T_1.
-        // N.B. w seems to be unused in the algorithm.
+        // N.B. w seems to be unused in the algorithm and y is not needed in
```
ltratt: So I have to ask: why is `w` defined, do you think? Is it an accident or ...?

tests/test_edit_script.rs
```diff
@@ -66,7 +79,21 @@ fn test_empty_arena() {
...
+    let ast_from = parse_file("tests/empty.calc", &get_lexer("grammars/calc.l"), &get_parser("grammars/calc.y")).unwrap();
+    let ast_to = Arena::new();
+    // Generate mappings between ASTs.
+    let matcher_config = MyersConfig::new();
+    let store = matcher_config.match_trees(ast_from, ast_to);
+    assert!(store.from.borrow().is_empty());
+    assert!(store.to.borrow().is_empty());
+    // Generate an edit script.
+    let gen_config: Box<EditScriptGenerator<String>> = Box::new(Chawathe96Config::new());
+    let _edit_script_wrapped = gen_config.generate_script(&store); // Panic.
```
ltratt: What's the point of the `let` here?

tests/common.rs
```diff
@@ -83,5 +83,17 @@ pub fn check_files(path1: &str, path2: &str, matcher: Box<MatchTrees<String>>) {
                "Final mapping not total for files: {} and {}.",
                path1, path2);
     // Test 2: final from and to ASTs should be isomorphic.
-    assert!(store.is_isomorphic(root_from, root_to));
+    // Isomorphism is defined strictly, as there being no differences in the two
```
ltratt: Remove the comma after "strictly"?

src/lib/edit_script.rs
```diff
@@ -45,7 +45,11 @@ use action::{ApplyAction, Delete, EditScript, Insert, Move, Update};
 use ast::{ArenaError, FromNodeId, NodeId, ToNodeId};
 use matchers::{EditScriptResult, MappingStore, MappingType};
 
-const TMP_ROOT: &str = "_____DIFFRACT_TMP_ROOT";
+/// Label to use when introducing temporary root nodes in an AST.
+///
+/// Temporary root nodes should be removed before the edit script generator
+/// completes. This constant is exposed for use in integration tests.
+pub const TMP_ROOT: &str = "_____DIFFRACT_TMP_ROOT";
```
ltratt: Any reason for quite so many underscores? ;)

ltratt:
So I must admit, this stuff is beyond my ken. I only have very shallow comments.

src/lib/edit_script.rs
snim2: I think it's a typo. `w` is used in the other two functions / methods so I would guess it's a copy & paste error.

ltratt: Is it a parameter to the function or ... ?

src/lib/edit_script.rs
snim2: No clever reason, I wanted to be reasonably sure that we didn't clash with any other label that might have been used in the tree.

src/lib/edit_script.rs
snim2: No, it's just a variable. Most of the comments in this code are lifted directly from the algorithm in the paper, which says `1. Let y=p(x) in T_2 and let w be the partner of x (x in T_1).`. In this notation `p(x)` means "parent of `x`". The other steps do not refer to `w` at all.

snim2:
(That last commit will need squashing)

src/lib/edit_script.rs
ltratt: We're fully in control of all the labels, right? Can any of them start with an underscore?

src/lib/edit_script.rs
ltratt: So do you think `w` is a typo for another variable or is the whole "let `w` be the partrner of x" bit redundant?

snim2: I think that phrase is redundant, all the other variables are clearly defined and seem perfectly sensible!

src/lib/edit_script.rs
snim2: Well, the labels come from the text of the original program, so they can usually start with an underscore (unless the lexer objects, I guess). In the grammars-with-whitespace  you introduce new node types `^~` and `~` to deal with these situations, but I didn't want to add anything special to the grammars (and not all of our grammars have whitespace nodes), so the "temporary root nodes" have a default type (i.e. the type is made by saying `Default::default()`) and this weird label.

src/lib/edit_script.rs
ltratt: OK, so could you expand the comment to make this clearer for worriers like me? :)

src/lib/edit_script.rs
ltratt: What I suggest is that you put a prefix on all nodes outside your control, so you can guarantee there can never be clashes. Or would that involve changing too many things?

snim2: Well, that means iterating over the whole tree twice to add the prefixes and then remove them, which seems a bit much. I'm not sure whether there is a better solution?

tests/test_edit_script.rs
```diff
@@ -90,7 +90,7 @@ fn test_to_arena_empty() {
     assert!(store.to.borrow().is_empty());
     // Generate an edit script.
     let gen_config: Box<EditScriptGenerator<String>> = Box::new(Chawathe96Config::new());
-    let _edit_script_wrapped = gen_config.generate_script(&store); // Panic.
+    gen_config.generate_script(&store); // Panic.
```
snim2: Of course, this gives a Rust warning `unused std::result::Result which must be used`

ltratt: You can use `.ok()` or `.unwrap()`.

src/lib/edit_script.rs
ltratt: Well, how about whenever you create a node you at least check it doesn't clash with that special ID? Basically I have learnt from bitter experience that unprotected magic IDs eventually blow up in my face one day...

snim2:
OK, I think these last commits answer your issue with the fake AST roots. I'm running the ignored integration tests locally now, just in case.

ltratt:
This looks good to me. I think it's time to squash?

snim2:
OK, squashed. I'd quite like to leave the fixes separate like this, because the commit messages make a bit more sense this way (I think).

ltratt:
Works for me.

PR #87, opened 2018-04-03
Decorate ASTs with hashes of each node (needed by GumTree).

snim2:
The fingerprinting module implements the algorithm described in M. Chilowicz, E. Duris, G. Roussel (2009) Syntax Tree Fingerprinting: A Foundation For Source Code Similarity Detection. This is used by the GumTree matcher algorithm.
    
The implementation here is based closely on the GumTree implementation. 

Implement [to_short_string()](https://github.com/GumTreeDiff/gumtree/blob/066d138201acb3c457bafb131a5a33fd820d1919/core/src/main/java/com/github/gumtreediff/tree/AbstractTree.java#L245) and [to_static_hash_string()](https://github.com/GumTreeDiff/gumtree/blob/066d138201acb3c457bafb131a5a33fd820d1919/core/src/main/java/com/github/gumtreediff/tree/AbstractTree.java#L228) on `NodeId`. Compute [tree isomorphism based on hashes](https://github.com/GumTreeDiff/gumtree/blob/066d138201acb3c457bafb131a5a33fd820d1919/core/src/main/java/com/github/gumtreediff/tree/AbstractTree.java#L124) (as a method on matcher objects). The hash generator here is based on the default GumTree hash generator, i.e. the [Md5RollingHashGenerator](https://github.com/GumTreeDiff/gumtree/blob/develop/core/src/main/java/com/github/gumtreediff/tree/hash/RollingHashGenerator.java#L68) class.
    
In addition, we create a new enum in ast.rs, `FromOrToNodeId`, which can be used by functions and methods which need to match on either a `NodeId<FromNodeId>` or a `NodeId<ToNodeId>` type.

Integration tests check tree isomorphism based on hashes, and fingerprints are added to ASTs only if the user has selected the GumTree matcher.

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -42,6 +42,7 @@ use std::fmt;
 use std::marker::PhantomData;
 use std::ops::{Index, IndexMut};
 
+use fingerprint::{CLOSE_SYMBOL, HashType, OPEN_SYMBOL, SEPARATE_SYMBOL};
 use hqueue::HeightQueue;
 
 /// Errors raised by arenas.
@@ -69,6 +70,15 @@ pub struct FromNodeId;
 /// A node identifier for a 'to' `Arena`.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd)]
 pub struct ToNodeId;
+
+/// A type which can either be `FromNodeId` or `ToNodeId`.
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub enum FromOrToNodeId {
+    /// Id for a `from` node.
+    FromNodeId(NodeId<FromNodeId>),
+    /// Id for a `to` node.
+    ToNodeId(NodeId<ToNodeId>),
+}
 
 /// A node identifier used for indexing nodes within a particular `Arena`.
 ///
@@ -320,6 +330,10 @@ pub struct Node<T: Clone, U: PartialEq + Copy> {
     ///
     /// Note that nodes with a non-`None` index are unique.
     index: Option<NodeId<U>>,
+    /// A hash that identifies this node.
+    ///
+    /// Used by the GumTree algorithm, and potentially to check tree isomorphism.
+    hash: Option<HashType>,
     /// The `label` of a node is the lexeme it corresponds to in the code.
     ///
     /// For example, a numerical or string literal, bracket, semicolon, etc.
@@ -362,7 +376,8 @@ impl<T: Clone> From<Node<T, ToNodeId>> for Node<T, FromNodeId> {
                                 col_no: node.col_no,
                                 line_no: node.line_no,
                                 char_no: node.char_no,
-                                token_len: node.token_len, }
+                                token_len: node.token_len,
+                                hash: node.hash, }
     }
 }
 
@@ -380,7 +395,8 @@ impl<T: Clone> From<Node<T, FromNodeId>> for Node<T, ToNodeId> {
                               col_no: node.col_no,
                               line_no: node.line_no,
                               char_no: node.char_no,
-                              token_len: node.token_len, }
+                              token_len: node.token_len,
+                              hash: node.hash, }
     }
 }
 
@@ -404,7 +420,8 @@ impl<T: Clone, U: PartialEq + Copy> Node<T, U> {
                col_no,
                line_no,
                char_no,
-               token_len, }
+               token_len,
+               hash: None, }
     }
 
     /// Return the Id of the parent node, if there is one.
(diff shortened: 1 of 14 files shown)


src/lib/fingerprint.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+fn out_seed<T: Clone + Eq + Hash + ToString, U: PartialEq + Copy>(id: &NodeId<U>,
+                                                                  arena: &Arena<T, U>)
+                                                                  -> String {
+    arena[*id].ty.to_string() + SEPARATE_SYMBOL + &arena[*id].label + CLOSE_SYMBOL
+}
+
+/// Convert a byte array (from digest) to integer.
+fn bytes_to_int(b: &[u8]) -> i64 {
```
ltratt: `u8` -> `i64` feels a bit weird. I wonder if this should be `u64`?

src/lib/ast.rs
```diff
@@ -513,6 +540,20 @@ impl<U: PartialEq + Copy> NodeId<U> {
         1 + self.descendants(arena).count() as u32
     }
 
+    /// Return a static hash of this node (used to compute isomorphism).
```
ltratt: What's a static hash?

src/lib/fingerprint.rs
snim2: Quite possibly, this is a Java `int` in the GT code

src/lib/fingerprint.rs
```diff
@@ -0,0 +1,235 @@
...
+    debug_assert_eq!(4, b.len(), "Byte array from digest has length {}", b.len());
+    i64::from(b[3]) & 0xff | (i64::from(b[2]) & 0xff) << 8 | (i64::from(b[1]) & 0xff) << 16
+    | (i64::from(b[0]) & 0xff) << 24
+}
+
+/// Integer exponent function, returns `b**e`.
+fn fpow(b: i64, e: i64) -> i64 {
```
ltratt: Is there not a builtin function (or a crate) for this?

snim2:
Please check that last commit!

src/lib/fingerprint.rs
```diff
@@ -0,0 +1,235 @@
...
+            FromOrToNodeId::FromNodeId(id) => {
+                let mut size = i64::from(id.size(&store.from_arena.borrow()));
+                let mut hash = HashGenerator::<T>::hash(self, &in_seed(&id, &store.from_arena.borrow()))
+                               * fpow(BASE, size);
+                for child in id.children(&store.from_arena.borrow()) {
+                    size -= i64::from(child.size(&store.from_arena.borrow())) * 2;
+                    debug_assert!(child.get_hash(&store.from_arena.borrow()).is_some());
+                    hash += child.get_hash(&store.from_arena.borrow()).unwrap()
+                }
+                hash += HashGenerator::<T>::hash(self, &out_seed(&id, &store.from_arena.borrow()));
```
ltratt: This line can be simplified to `hash + HashGenerator::...` (no semicolon) and the line below it deleted.

src/lib/fingerprint.rs
```diff
@@ -0,0 +1,235 @@
...
+            FromOrToNodeId::ToNodeId(id) => {
+                let mut size = i64::from(id.size(&store.to_arena.borrow()));
+                let mut hash = HashGenerator::<T>::hash(self, &in_seed(&id, &store.to_arena.borrow()))
+                               * fpow(BASE, size);
+                for child in id.children(&store.to_arena.borrow()) {
+                    size -= i64::from(child.size(&store.to_arena.borrow())) * 2;
+                    debug_assert!(child.get_hash(&store.to_arena.borrow()).is_some());
+                    hash += child.get_hash(&store.to_arena.borrow()).unwrap()
+                }
+                hash += HashGenerator::<T>::hash(self, &out_seed(&id, &store.to_arena.borrow()));
```
ltratt: Same comment as above.

src/lib/matchers.rs
```diff
@@ -163,6 +163,14 @@ impl<T: Clone + Debug + Eq + 'static> MappingStore<T> {
         self.get_to(from).map_or(false, |x| x == *to)
     }
 
+    /// Compute whether two sub-trees are isomorphic, based on their hashes.
+    ///
+    /// Isomorphism in this case is defined has having the same hash and the
+    /// same hash serialization.
+    pub fn is_isomorphic_hash(&self, from: NodeId<FromNodeId>, to: NodeId<ToNodeId>) -> bool {
```
ltratt: I assume we only use this function in places where we want check whether two things *aren't* equal? Even then, two things which hash to the same value might be distinct. So I'm a bit nervous about this function.

snim2: It seems odd, I agree, but this is as specified in the GT code:

```java
    @Override
    public boolean isIsomorphicTo(ITree tree) {
        if (this.getHash() != tree.getHash())
            return false;
        else
            return this.toStaticHashString().equals(tree.toStaticHashString());
    }
```

ltratt: I'm about 99.9% certain that code is incorrect and works only because hash collisions are somewhat rare. But relying on hashes for correctness is a bad idea: they're an optimisation, not a guarantee.

snim2: I don't disagree (and we already have an exact isomorphism test that we use in integration tests), but this is the way GT does it, so I tend to think we should start with this and see what happens in practice, rather than deviating from their implementation before we can do any testing. I'll raise an issue.

ltratt: Can we at least add a `debug_assert` which fails if a) the hashes are the same but b) the contents aren't? Otherwise this will get lost and forgotten until one day it bites us on the rear end.

snim2:
OK, try that last assertion...

ltratt:
Yep!

snim2:
Squashed

PR #94, opened 2018-04-18
q-gram distance

snim2:
Calculate a q-gram distance between strings. Needed by the Zhang-Shasha matcher.

Implementation based on the simmetrics library: https://github.com/Simmetrics/simmetrics

### DIFF
--- a/src/lib/qgram.rs
+++ b/src/lib/qgram.rs
@@ -1,0 +1,284 @@
+// Copyright (c) 2018 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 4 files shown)


snim2:
Ah, hang about, there are some clippy suggestions to go in. Sorry!

snim2:
OK, will need squashing

ltratt:
Do you want to squash this before I review?

snim2:
Squashed!

src/lib/qgram.rs
```diff
@@ -0,0 +1,235 @@
...
+            return token_set;
+        } else if padded.len() < self.q as usize {
+            token_set.insert(padded);
+            return token_set;
+        }
+        // To parse a string into a q-gram set of tokens we move a `self.q`-wide
+        // sliding windows across the string.
+        let chars = padded.chars().collect::<Vec<char>>();
+        let last_qgram_index = chars.len() - self.q as usize;
+        for qgram_start in 0..=last_qgram_index {
```
ltratt: Stray `=`? Or maybe it does something I don't know about?

ltratt:
The whole `#` thing seems weird to me -- is there a justification for this in the literature? I'd be curious to have a cut-down explanation!

src/lib/qgram.rs
snim2: This is equivalent to `for (qgram_start = 0; i <= last_qgram_index; i++) { ... }` so the `..=` is like the Java `<=`.

snim2:
So, when we perform a q-gram distance computation, we first tokenise the input string. If the string is `ABCDEF` then (without padding) the 3-gram tokenisation would be `ABC, BCD, CDE, DEF`. Notice that `C` appears three times in that list, but not so `A` or `F`. With padding, we would tokenise to the list `##A, #AB, ABC, BCD, CDE, DEF, EF#, F##` and every letter appears 3 times. Sometimes people use `^` and `$` for the start and end padding tokens, but we should probably use a token that we believe is less likely to appear in string literals and comments. I'm not sure that `#` is the best choice, but it was the default padding string in the simmetrics library that GT uses.

src/lib/qgram.rs
ltratt: I didn't know that. I don't like it either (Rust, why so much syntax?!) but fair enough.

ltratt:
What happens then if the user string *does* contain `#`? Does it throw the algorithm off a little or ... ?

src/lib/qgram.rs
vext01: Gross!

ltratt:
"Same result" being a boolean or ... ?

snim2:
No, the distance is a float

ltratt:
Hmm.

ltratt:
Do we only ever compare pairs of strings at a time? If so, we could use a "fresh" character: i.e. a character `c` which is not in either of the two strings. With unicode this should be pretty easy to do.

snim2:
Indeed we do only compare pairs of strings. I'm not sure what you mean about unicode though?

ltratt:
Well, in theory, someone could give us two strings that between them use every possible unicode character. But those would be two long strings (4GiB between them), so I think we can assume that won't happen. What that means is that for two strings S1 and S2, it should be easy to find a character C that doesn't appear in either.

ltratt:
Ping @snim2.

snim2:
OK, this is ready for review. These last commits:

  * Default to a padding character that is highly unlikely to appear in real strings (the "replacement" char which is used to indicate a unicode decoding error)
  * If the default padding character does appear in a real string, we look for an alternative.

src/lib/qgram.rs
```diff
@@ -83,15 +83,44 @@ impl QGram {
         distance
     }
 
+    /// Choose a padding character that does not appear in `input`.
+    ///
+    /// Use `self.padding` if possible.
+    /// Panics if it cannot find a suitable padding character.
+    fn choose_padding(&self, input: &str) -> String {
+        // Note that `first()` is not defined for the `Chars` type.
```
ltratt: Comment is unneeded.

src/lib/qgram.rs
```diff
@@ -83,15 +83,44 @@ impl QGram {
...
+        if !input.contains(self.padding.chars().nth(0).unwrap()) {
```
ltratt: This makes this function O(2n) when we can make this O(n) by folding this into the `for` loop below and having a flag `found_padding_char` which is set to true if `c == padding_char`. [Hopefully that makes sense.]

snim2:
OK, hopefully that addresses the comments.

ltratt:
OK, I think we're ready to squash and rebase against master.

ltratt:
@snim2 Is this ready?

snim2:
Sorry, for got to say -- ready and waiting!

PR #95, opened 2018-05-08
Code cleanup

snim2:
Various improvements suggested by the compiler and clippy. Unfortunately there is a lot of this, and a `rustfmt` commit in the middle, so it won't be the easiest to review.

The first few commits are from various parts of the code base, followed by a number from the Chawathe 98 matcher code. I have ignored clippy suggestions on cyclomatic complexity and functions with too many arguments (some of the suggestions might be sensible, but they require more major surgery).

Needs much squashing.

### DIFF
--- a/src/lib/chawathe98_matcher.rs
+++ b/src/lib/chawathe98_matcher.rs
@@ -2398,8 +2398,6 @@ pub trait MatchingTreesScriptor<T: Clone + Debug> {
 pub fn chawathe_matching_actual(
     mut from_arena: Arena<String, FromNodeId>,
     mut to_arena: Arena<String, ToNodeId>,
-    from_path: &str,
-    to_path: &str,
     user_defined_cost: CostEdge,
 ) -> EditScriptResult<String> {
     from_arena.new_node(
--- a/src/lib/chawathe98_matcher.rs
+++ b/src/lib/chawathe98_matcher.rs
@@ -43,18 +43,18 @@
 /// This matcher would use the additional operation which are GLUE and COPY.
 /// In addition to Insert, Delete, Move, Update.
 /// This is supposed to be an alternative matcher to chawathe96.
-use std::fmt::Debug;
-
-use ast::{Arena, FromNodeId, NodeId, ToNodeId};
-use matchers::{has_same_type, has_same_type_and_label};
 
 use action::{ApplyAction, Copy, Delete, EditScript, Glue, Insert, Move, Update};
+use ast::{Arena, FromNodeId, NodeId, ToNodeId};
 use emitters::RenderJson;
-use matchers::EditScriptResult;
+use matchers::{has_same_type, has_same_type_and_label, EditScriptResult};
+
 use std::cell::RefCell;
 use std::cmp;
 use std::collections::HashMap;
 use std::collections::HashSet;
+use std::fmt::Debug;
+use std::hash::BuildHasher;
 
 /// Not needed by matching algorithms, but useful for debugging.
 #[derive(Debug, Clone, Eq, Hash, PartialEq, Copy)]
@@ -76,8 +76,9 @@ pub enum EdgeType {
     /// Edge corresponds to a OK operation.
     OK,
 }
+
+/// The initial cost of each edge
 #[derive(Debug, Clone)]
-/// The intial cost of each edge
 pub struct CostEdge {
     /// Edge corresponds to a UPDATE cost operation.
     pub update: usize,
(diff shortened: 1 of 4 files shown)


src/lib/null_matcher.rs
```diff
@@ -49,13 +49,13 @@ use ast::{Arena, FromNodeId, ToNodeId};
...
 /// The Null matcher needs no configuration.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Default, Clone, PartialEq)]
 pub struct NullConfig {}
 
 impl NullConfig {
     /// Create a new configuration object, with default values.
     pub fn new() -> NullConfig {
-        NullConfig {}
+        Default::default()
```
ltratt: Do we need this? Because, IMHO, the result is not easier to read...

src/main.rs
```diff
@@ -414,32 +414,21 @@ fn main() {
...
-                chawathe98_matcher::chawathe_matching_actual(ast_base_clone,
-                                                             ast_diff_clone,
-                                                             &args.arg_base_file,
-                                                             &args.arg_diff_file,
-                                                             cost_user_default);
-            }
+                chawathe98_matcher::CostEdge::new(1, 1, 1, 1, 1, 1, 0, 0)
+            };
+            // Chawathe98 matcher does not yet have a way to generate a textual diff.
+            let _ =
```
ltratt: This assignment is pointless, right?

src/lib/chawathe98_matcher.rs
```diff
@@ -681,9 +681,9 @@ pub fn edgecover_solver(
     // We would check through the edge keep and only retain edges in edge keep;
     let mut edge_filtered: Vec<Edge> = Vec::new();
     for edge in new_matcher_pruning.list_edges.borrow().iter() {
-        if edge_keep.contains(&(edge.from_node, edge.to_node, edge.edge_type, edge.value)) {
-            edge_filtered.push(Edge::new(edge.from_node, edge.to_node, edge.value, edge.edge_type));
-        } else if edge.edge_type == EdgeType::DELETE {
+        if edge_keep.contains(&(edge.from_node, edge.to_node, edge.edge_type, edge.value))
+           || edge.edge_type == EdgeType::DELETE
```
ltratt: Wouldn't it be worth having the simple case first? I don't know how LLVM chooses to order these, but it might choose to keep the order unchanges -- and the boolean check is bound to be faster.

src/lib/chawathe98_matcher.rs
```diff
@@ -1534,46 +1534,31 @@ pub fn edgecover_apply_edit_script(new_matcher_pruning: &mut MappingStoreGraph<S
...
-                            insert_action.apply(&mut from_arena)
-                                         .expect("Cannot apply INSERT action to AST.");
-                            script.push(insert_action);
-
-                            indexes_t_f.insert(to_node.id(), new_node.id());
-                        } else {
-                            if get_position.unwrap() > size_of_children {
-                                get_position = Some(size_of_children);
-                            }
+                        // Note, we used to check: if get_position.unwrap() != 0
```
ltratt: Is this comment important?

src/lib/chawathe98_matcher.rs
```diff
@@ -1534,46 +1534,31 @@ pub fn edgecover_apply_edit_script(new_matcher_pruning: &mut MappingStoreGraph<S
...
-                                                                Some(from_node_parent),
-                                                                get_position.unwrap() as u16);
-                            // Parent exists
-                            insert_action.apply(&mut from_arena)
-                                         .expect("Cannot apply INSERT action to AST.");
-                            script.push(insert_action);
+                        let mut insert_action = Insert::new(new_node,
+                                                            Some(from_node_parent),
+                                                            get_position.unwrap() as u16);
+                        // Parent exists
```
ltratt: This comment doesn't really make sense to me. It sounds like it's a comment about a conditional, but there is no conditional here...

src/lib/chawathe98_matcher.rs
```diff
@@ -2086,16 +2066,18 @@ impl<T: Clone + Debug + Eq + 'static> MatchingTreesScriptor<T> for Config2 {
 
             // Lets get edge E2
             let e2_to_node_checker = helper_function.1.get(&m);
-            let mut e2_to_node = None;
-            if e2_to_node_checker.is_some() {
-                e2_to_node = e2_to_node_checker.unwrap().0.iter().nth(0);
-            }
+            let e2_to_node = if e2_to_node_checker.is_some() {
```
ltratt: Could this be simplified to something like `let e2_to_node = e2_to_node_checker.map(|x| x.0.first());`?

src/lib/chawathe98_matcher.rs
```diff
@@ -2086,16 +2066,18 @@ impl<T: Clone + Debug + Eq + 'static> MatchingTreesScriptor<T> for Config2 {
...
+            } else {
+                None
+            };
 
             let e3_from_node_checker = helper_function.0.get(&n);
-            let mut e3_from_node = None;
-            if e3_from_node_checker.is_some() {
-                e3_from_node = e3_from_node_checker.unwrap().0.iter().nth(0);
-            }
+            let e3_from_node = if e3_from_node_checker.is_some() {
```
ltratt: Same comment as above.

ltratt:
A few minor comments. And, to check, are you running rustfmt on every commit these days? I'd like to avoid jumbo rustfmt commits, as they ruin history, and make reviewing harder.

snim2:
I am running `rustfmt`. Unfortunately it fails if a file has overly long lines, and because I'm running it in my text editor, I don't see the failure. That's the reason for the massive commit in this PR, hopefully it won't happen again.

snim2:
OK, I think that answers all the comments. Hopefully this is ready to squash.

ltratt:
Please squash.

snim2:
Squashed.

snim2:
Sorry, one of those commits has a spelling error...

snim2:
OK, typo fixed.

PR #97, opened 2018-05-18
Refactor unit tests

snim2:
This PR:

 * Factors out common ASTs used in many unit tests across the code base.
 * Provides an API for loading ASTs from XML in test config. 
 * Converts the common parse trees to XML.

The reason for doing this, other than avoiding code duplication, is to cut down on the amount of code in the unit tests, which is currently large and broadly unreadable. Also, this will make it easier in other PRs to load XML test data similar to that in the GT code base.

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -102,7 +102,10 @@ pub trait Patchify<T: Clone + fmt::Debug + ToString> {
     /// Turn object into a `Patch`. Non-terminal nodes are ignored.
     fn patchify(&self, store: &MappingStore<T>, from: &mut Vec<Patch>, to: &mut Vec<Patch>);
     /// Turn object into a 'Patch'.
-    fn patchify_chawathe98(&self, store: &MappingStoreGraph<T>, from: &mut Vec<Patch>, to: &mut Vec<Patch>);
+    fn patchify_chawathe98(&self,
+                           store: &MappingStoreGraph<T>,
+                           from: &mut Vec<Patch>,
+                           to: &mut Vec<Patch>);
 }
 
 /// A list of actions to be applied.
@@ -213,7 +216,7 @@ impl Glue {
     pub fn new(from: NodeId<FromNodeId>, parent: NodeId<FromNodeId>, position: u16) -> Glue {
         Glue { from_node: from,
                node: parent,
-               pos : position}
+               pos: position, }
     }
 }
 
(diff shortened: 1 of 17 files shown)


tests/test_edit_script.rs
```diff
@@ -51,46 +51,40 @@ use common::check_files;
...
-    // Generate mappings between ASTs.
     let matcher_config = MyersConfig::new();
     let store = matcher_config.match_trees(ast_from, ast_to);
     assert!(store.from.borrow().is_empty());
     assert!(store.to.borrow().is_empty());
-    // Generate an edit script.
-    let gen_config: Box<EditScriptGenerator<String>> = Box::new(Chawathe96Config::new());
-    let _edit_script_wrapped = gen_config.generate_script(&store); // Panic.
+    Box::new(Chawathe96Config::new()).generate_script(&store)
+                                     .ok(); // Panic.
```
ltratt: `ok` doesn't panic?

src/lib/action.rs
```diff
@@ -213,7 +216,7 @@ impl Glue {
     pub fn new(from: NodeId<FromNodeId>, parent: NodeId<FromNodeId>, position: u16) -> Glue {
         Glue { from_node: from,
                node: parent,
-               pos : position}
+               pos: position, }
```
ltratt: The `, ` seems unnecessary?

src/lib/mod.rs
```diff
@@ -114,7 +122,65 @@ pub mod fingerprint;
...
-    use ast::{Arena, FromNodeId};
+    use serde::{Deserialize, Deserializer};
+    use serde_xml_rs::deserialize;
+
+    use ast::{Arena, FromNodeId, NodeId};
+
+    fn to_string<'de, D>(deserializer: D) -> Result<String, D::Error>
+        where D: Deserializer<'de>
+    {
+        let s: String = Deserialize::deserialize(deserializer)?;
```
ltratt: Could this just be `Ok(Deserialize::deserialize(deserializer)?)`?

tests/test_edit_script.rs
snim2: It does panic, yep

src/lib/action.rs
snim2: Added by rustfmt!

snim2:
OK to squash?

tests/test_edit_script.rs
ltratt: What's the return type of `generate_script`? `Result::ok()` says "Converts self into an Option<T>, consuming self, and discarding the error, if any."

src/lib/action.rs
ltratt: Oh god, can we turn that feature off? I mean, that's just ugly...

tests/test_edit_script.rs
snim2: I'm not sure that I follow this. The return type of `generate_script()` is `EditScriptResult<T>` which is an alias for `Result<EditScript<T>, ArenaError>`. The `ok()` swallows the `ArenaError`, but it doesn't effect any `panic!()`s. If we replace lines 73 and 74 above with a call to the `panic!` macro, the results of the unit tests don't change. 

Have I missed something?

ltratt: I think we're partly agreeing, but I think we should remove the text that says `panic` because it's not clear why it says that (to me, at least).

snim2: I think I just put them in to make clear which LOC causes the panic.

snim2:
OK to squash?

ltratt:
Please squash.

snim2:
Squashed!

PR #98, opened 2018-05-24
Small api changes

snim2:
Some small API changes needed by the ZS work.

### DIFF
--- a/src/lib/ast.rs
+++ b/src/lib/ast.rs
@@ -734,6 +734,11 @@ impl<U: PartialEq + Copy> NodeId<U> {
         Ok(())
     }
 
+    /// Return the nth child of this node, if one exists.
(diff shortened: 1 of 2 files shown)


src/lib/ast.rs
```diff
@@ -254,6 +254,18 @@ impl<T: Clone + PartialEq, U: PartialEq + Copy> Arena<T, U> {
         false
     }
 
+    /// Return the id of the first leaf node in this arena, if there is one.
+    pub fn get_first_leaf(&self) -> Option<NodeId<U>> {
+        if self.is_empty() {
```
ltratt: It would probably be more idiomatic to do something like:

```rust
if let Some(node) = self.root {
   ...
} else {
   None
}
```

snim2:
I did not know about the `mut` in  `let Some(mut var) = val { ... } else { ... }` before. 

Ready to squash?

ltratt:
Yep, please squash.

snim2:
Squashed

PR #99, opened 2018-06-06
Fix serde versions.

snim2:
These versions are chosen to work with the snapshot of rust we
need to use.


ltratt:
I can confirm this fixes the compilation problems. I wouldn't have thought of this, so well done!

snim2:
Hang about though -- the tests fail! I think that's because of the latest changes in lrlex, I'll try to fix in this PR.

ltratt:
All passing -- merge as-is?

snim2:
Yes please!

PR #100, opened 2018-06-07
Prefix lines of pretty-printed ASTs with NodeIds.

snim2:
This is intended to help with debugging.
As an example, this AST:
```
^~"
  "~"
  "Expr"
    "Term"
      "Factor"
        "INT" 1
```
becomes:
```
000: "^~"
001:   "~"
002:   "Expr"
003:     "Term"
004:       "Factor"
005:         "INT" 1
```

### DIFF
--- a/src/lib/action.rs
+++ b/src/lib/action.rs
@@ -648,11 +648,11 @@ mod test {
     #[test]
     fn apply_delete_leaf() {
         let mut arena = create_mult_arena();
-        let format1 = "\"Expr\" +
-  \"INT\" 1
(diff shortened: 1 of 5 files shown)


ltratt:
Is the "00" thing useful? I mean, I can imagine it working less well after 999 ;) If you know how many lines you're going to see, you could pad things? Or maybe don't pad?

snim2:
It means that the white space next to the node id is always "correct". If I need to debug an AST of >=1000 lines I think I'll have more problems than integer padding!

PR #101, opened 2018-06-08
Implementation of the Zhang Shasha (1989) paper

snim2:


ltratt:
Can you give a description of this, including a rough overview of what's changed and why?

snim2:
This PR implements the matching algorithm described in Zhang Shasha (1989). We need this algorithm to implement GumTree, because GT uses ZS as a fall-back in its "bottom-up" matcher. The PR includes unit tests for the ZS matcher, and adds the matcher to the diffract front-end:

```
$ diffract_run.sh --list
...

Zhang-Shasha
-------
The Zhang Shasha tree matching algorithm -- works by attempting to match
successive subtrees. For more information see Zhang and Shasha (1989) Simple
Fast Algorithms for the Editing Distance Between Trees and Related Problems.
```

...so the new matcher can be used by passing `-m ZhangShasha` to diffract on the command line.

src/lib/ast.rs
```diff
@@ -762,6 +750,15 @@ impl<U: PartialEq + Copy> NodeId<U> {
                    .position(|val| val == self)
     }
 
+    /// Return the id of the first leaf of the subtree rooted on this node, if there is one.
+    pub fn get_first_leaf<T: Clone>(self, arena: &Arena<T, U>) -> NodeId<U> {
+        let mut node = self;
+        while !node.is_leaf(arena) {
+            node = node.children(&arena).collect::<Vec<NodeId<U>>>()[0]
```
ltratt: If the first `node.is_lef()` call is `false`, doing a `collect` here is going to be pretty inefficient: it has to malloc memory, copy things over etc. etc. Can we not do a version that simply references things rather than copying?

src/lib/zs_matcher.rs
```diff
@@ -37,27 +37,104 @@
...
+                            dst: ZSTree::new(dst),
+                            tree_dist: vec![vec![]],
+                            forest_dist: vec![vec![]] }
+    }
+
+    /// Compute the tree distance matrix. See Section 3.2 of the paper.
+    fn compute_tree_distance<T: Clone + Debug + Eq + ToString + 'static>(&mut self,
+                                                                         store: &MappingStore<T>)
+    {
+        self.tree_dist =
```
ltratt: Couldn't these two assignments be put in the `for` loop immediately afterwards? That would be easier to read and, probably, faster.

src/lib/zs_matcher.rs
```diff
@@ -37,27 +37,104 @@
...
+                                        .collect();
+        self.forest_dist =
+            (0..self.src.node_count + 1).map(|_| Vec::with_capacity(self.dst.node_count + 1))
+                                        .collect();
+        for i in 0..self.src.node_count + 1 {
+            for _ in 0..self.dst.node_count + 1 {
+                self.tree_dist[i].push(0.0);
+            }
+        }
+        for i in 0..self.src.node_count + 1 {
```
ltratt: Can't this `for` loop be folded into its predecessor?

src/lib/zs_matcher.rs
```diff
@@ -37,27 +37,104 @@
...
+        }
+    }
+
+    /// Compute the forest distance matrix. See Section 3.2 of the paper.
+    fn forest_dist<T: Clone + Debug + Eq + ToString + 'static>(&mut self,
+                                                               i: usize,
+                                                               j: usize,
+                                                               store: &MappingStore<T>) {
+        self.forest_dist[self.src.lld(i) - 1][self.dst.lld(j) - 1] = 0.0;
+        for di in self.src.lld(i)..(i + 1) {
```
ltratt: No need for the brackets around `i + 1`.

src/lib/zs_matcher.rs
```diff
@@ -37,27 +37,104 @@
...
+            let cost_del = get_deletion_cost(&self.src.id(di), &store.from_arena.borrow());
+            self.forest_dist[di][self.dst.lld(j) - 1] =
+                self.forest_dist[di - 1][self.dst.lld(j) - 1] + cost_del;
+            for dj in self.dst.lld(j)..(j + 1) {
```
ltratt: Ditto brackets.

src/lib/zs_matcher.rs
ltratt: I don't know if the `self.src.lld(i)` call is expensive or not, but if it is, you could assign it to a variable as it's used quite often below?

src/lib/zs_matcher.rs
```diff
@@ -70,36 +147,361 @@ Fast Algorithms for the Editing Distance Between Trees and Related Problems.";
...
+                self.forest_dist(last_row, last_col, &store);
+            }
+            root_pair = false;
+            let first_row = self.src.lld(last_row) - 1;
+            let first_col = self.dst.lld(last_col) - 1;
+            let mut row = last_row;
+            let mut col = last_col;
+            while row > first_row || col > first_col {
+                if row > first_row
+                   && self.forest_dist[row - 1][col] + 1.0 == self.forest_dist[row][col]
```
ltratt: I get nervous about any floating point calculation that relies on `==` or `!=` -- are we sure this is doing what we think it is?

src/lib/zs_matcher.rs
```diff
@@ -70,36 +147,361 @@ Fast Algorithms for the Editing Distance Between Trees and Related Problems.";
...
+                    let src_node = self.src.id(row);
+                    let dst_node = self.dst.id(col);
+                    if has_same_type(&src_node,
+                                     &store.from_arena.borrow(),
+                                     &dst_node,
+                                     &store.to_arena.borrow())
+                    {
+                        store.push(src_node, dst_node, &MappingType::ANCHOR);
+                    } else {
+                        assert!(false, "Cannot map nodes with different types.");
```
ltratt: Why not:
```rust
assert!(!has_same_type(....));
store.push(...)
```
Although maybe that `assert` should be a `panic`?

src/lib/zs_matcher.rs
```diff
@@ -70,36 +147,361 @@ Fast Algorithms for the Editing Distance Between Trees and Related Problems.";
...
+                                llds: Vec::with_capacity(node_count),
+                                ids: Vec::with_capacity(node_count),
+                                key_roots: vec![] };
+        for _ in 0..node_count {
+            tree.llds.push(0);
+            tree.ids.push(NodeId::new(0));
+        }
+        debug_assert_eq!(tree.llds.len(), node_count);
+        debug_assert_eq!(tree.ids.len(), node_count);
+        let mut index = 1;
```
ltratt: It's generally more Rust-y to use `idx`.

src/lib/zs_matcher.rs
snim2: IIWYM, although this is taken quite directly from the GT code. Perhaps an epsilon would be better here?

snim2:
OK, I *think* that answers the comments...

src/lib/zs_matcher.rs
ltratt: IIWYM means ...?

We can't blindly add an epsilon without thinking about what's going on -- unfortunately there isn't going to be a substitute for thinking about what this is really doing. For example, how many decimal places might we be caring about? could/should we be converting this to integer arithmetic? and so on.

ltratt:
Apart from the FP comment (which, I admit, is a hard one) this is looking good.

src/lib/zs_matcher.rs
ltratt: If we use floats as the representation, then not even all integers are accurately represented. I don't think we can merge this until we have some answers to the questions I suggested earlier. I know this sucks, but if no-one else has been bothered thinking about this in the past, we might have to do so now...

snim2:
Re. floating-point. The matching algorithm is based on Algorithm Y from Wagner and Fischer (1974) which is a string-distance algorithm (originally for spelling checkers, I think). The first part of Z-S and W-F build a cost matrices, which show the cost of inserting / deleting / updating pair of subforests or prefix strings in the source / destination structures. Algorithm Y then navigates through the matrix from the bottom-right to the top-left cell to build a match or edit sequence. Like this:

```
start at bottom-right cell
until we reach the top-left cell
do
  1. if the value in the current cell == value in the cell to the left + deletion cost
      move to the cell to the left of the current cell
  2. if the value in the current cell == value in the cell below + insertion cost
      move to the cell below the current cell
  3. if the value in the upper-left cell == value in the current cell, record a match, 
      and move to the upper-left cell
end
```

The Z-S code is slightly more complicated than this, because it deals with forests rather than strings, so it has an outer loop which runs the W-F algorithm over each pair of forests and adds some extra checks to ensure that matches really are matching two trees. 

Where does this leave us with FP? If the check in 3. above is incorrect, we will miss a match and erroneously record an insertion or deletion. If the checks in 1. and 2. are incorrect we may miss an insert/delete and erroneously record a match (UPDATEs count as matches here). Because we have this outer loop over subforests, we may end up considering the "wrong" subforest next, which will result in extra or missed matches. So, the picture isn't as straight forward as tweaking the equality comparisons towards more or fewer matches.

A comparison against a DELTA (choosing the DELTA based on the QGram algorithm which we are using to compare labels in the ASTs) works with all the unit tests I have so far, I tend to think this is the best way forward, but I'm not sure that the argument here is very compelling.

ltratt:
So next question: how are the values calculated? If they're all (say) just of the form `X + N` then (I think) provided `X + N = Y` it doesn't matter if `X` and `N` are different (so `2 + 3 = 5` will compare equal to `4 + 1 = 5`). However, if there's a sequence of calculations, then the comparisons may fail.

The problem with the delta idea is that until we know the size of the error, we can't know what a correct delta is. Too big and we consider equal things that should be unequal; too small and we have the opposite problem.

snim2:
Each entry in the matrix is the minimum of the costs of a nearby cell + cost of insert / delete / update. The costs of an insertion or deletion are fixed at `1.0`. The cost of an update is more complicated:
```
if the two AST nodes are of the same type AND one or both are empty
    update_cost = 1.0
else if the two AST nodes are of the same type both are non-empty
    update_cost = 1.0 - trigram_distance(from_node, to_node);
else f64::MAX
```  

and `trigram_distance()` is the QGram distance algorithm we merged in #94. In the PR we used a delta in the unit tests, based on observing a number of test cases.

snim2:
I'm not sure about arbitrary, but there are some divisions and subtractions in there, it isn't just a matter of adding floats.

ltratt:
OK, so is there a *maximum* number of operations in `trigram_distance`? e.g. "at most 3 additions, 2 divisions, and 1 multiplication"? Or is it a potentially unbounded number of operations?

snim2:
The comparison is given by `1.0 - dist / string1.len() string2.len())` where `dist` is:
```
for token in union.distinct_elements() {
    distance += abs((string1.count_of(token) - string2.count_of(token)));
}
```

So, in `O(n)` terms it's bounded in some way by the lengths of the strings.

ltratt:
OK, so next question: is there a notion of the maximum possible error in FP for `n` operations (or, alternatively, of the likely error per operations)? If so, we might be able to make a (fairly accurate) dynamic delta.

snim2:
Ready for re-review (assuming the tests pass)

ltratt:
Looks good. Please squash as you see fit.

snim2:
Squashed

PR #107, opened 2018-07-11
Cleaner clippy output.

snim2:
When I run clippy, I get several screens of output that I'm not interested in, which makes it more likely that I'll miss something important. This PR suppresses some warnings that we do not care about.

Specifically:

  * Missing Default() traits.
  * High cyclomatic complexity and too many function arguments.
  * High type complexity.

### DIFF
--- a/src/lib/mod.rs
+++ b/src/lib/mod.rs
@@ -38,6 +38,11 @@
 #![feature(test)]
 #![feature(try_from)]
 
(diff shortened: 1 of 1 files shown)


snim2:
This PR needs bringing up to date with master, but otherwise it's good to go.

ltratt:
OK, please rebase first.

snim2:
Done

ltratt:
I completely agree with these!

PR #110, opened 2018-07-13
RTED -- second PR

snim2:
Add some getters and setters to the RTED module.

### DIFF
--- a/src/lib/rted.rs
+++ b/src/lib/rted.rs
@@ -36,6 +36,7 @@
 // SOFTWARE.
 
 #![warn(missing_docs)]
+#![allow(dead_code)]
 
 /// This matcher implements Pawlik and Augsten (2011).
 use std::fmt::Debug;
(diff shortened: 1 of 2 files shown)


src/lib/rted.rs
```diff
@@ -177,6 +178,56 @@ impl<'a> RTEDConfig<'a> {
...
+        }
+        if switched {
+            for i in 0..self.size_src {
+                for j in 0..self.size_dst {
+                    if self.itree_src.info[InfoIdx::Post2Size][i]
+                       >= self.itree_dst.info[InfoIdx::Post2Size][j]
+                    {
+                        self.strategy[i][j] = strategy;
+                    } else {
+                        self.strategy[i][j] = strategy + 4;
```
ltratt: Why `+ 4`? As you know, I really don't like magic constants ;)

snim2: Hrm. I'm not sure what the `4` means, but it looks like this method isn't called in the original code, so I guess we can rm it for now.

snim2:
Ready for re-review

src/lib/rted.rs
ltratt: I have to ask the obvious question: why did we port a function that isn't used? Does that mean our testing isn't catching this sort of thing?

snim2: The module is so big I'm splitting it into small PRs. This PR just adds some getters and setters that are not tested, but I had presumed they would be called by methods I haven't implemented yet. However, it looks like this one is just dead code in the Java version.

ltratt: When do tests get added? I'm not sure we're gaining much if we merge small PRs that don't have tests.

snim2: It's a good question. The original code has a very small number of tests which can really only be run when the whole algorithm is complete. For these set and get methods I'm not sure there is much that we can usefully do, but I think we can start creating more useful tests from the next PR onwards.

ltratt:
OK, if you rebase, I can merge.

PR #120, opened 2018-09-11
Change hqueue api

snim2:
This PR implements a number of changes in the hqueue (height queue) module which are needed for GumTree. 

Each commit makes a single change, so these will probably be easier to review separately, but should probably be squashed before merging.


ltratt:
This looks fine but, unless you're really sure, these commits look like they should remain separate to me. OK for me to merge?

snim2:
Yes, please do!

PR #122, opened 2018-09-19
Clippy suggestions

snim2:
From the latest version of clippy.


ltratt:
Do you want to squash?

snim2:
I think they probably make more sense separately.

